PR #4, opened 2022-01-25
Add basic action to build the package

sjavora:
Since there are no tests (yet?) it only tries to build the package, but that's good enough for now.


PavelHolec (approved):

hrach:
In repo settings may setup to auto-delete merged branches ^ :)

sjavora:
Thanks, enabled.

PR #5, opened 2022-01-26
ListChoice component

sjavora:


Sources/Orbit/Components/ListChoice.swift
```diff
@@ -0,0 +1,252 @@
...
+
+/// Shows one of a selectable list of items with similar structures.
+///
+/// - Note: [Orbit definition](https://orbit.kiwi/components/listchoice/)
+/// - Important: Component expands horizontally to infinity up to a ``Layout/readableMaxWidth``.
+public struct ListChoice: View {
+
+    let title: String
+    let description: String
+    let icon: Icon.Symbol
```
PavelHolec, line 11: For future reusability, we might probably need to use Icon.Content here, as some existing similar custom components already use flags in place of icon for example (and a client may need to provide a custom icon content).

Sources/Orbit/Components/ListChoice.swift
```diff
@@ -0,0 +1,252 @@
...
+            trailingElementView
+                .padding(.trailing, .medium)
+        }
+        .frame(maxWidth: Layout.readableMaxWidth, alignment: .leading)
+        .overlay(separator, alignment: .bottom)
+    }
+    
+    @ViewBuilder var header: some View {
+        if isHeaderEmpty == false {
+            HStack(alignment: .firstTextBaseline, spacing: .small) {
```
PavelHolec, line 49: It might be possible to use the `Header` component for this whole header. It has more correct and unified alignment of icon (plus auto-sizing of icon, based on title style).

The `Header` would need to support a non-Heading style, though. In this case, the large+medium text, rather than actual Heading style.

In case of not using the `Header` yet, the `Text(title)` can be simply updated to `Text(title, weight: .medium)`.

Sources/Orbit/Components/ListChoice.swift
```diff
@@ -0,0 +1,252 @@
...
+                iconView
```
PavelHolec, line 50: Based on Figma, the spacing of icon is ~12 leading and ~8 trailing around the icon. To avoid custom dynamic paddings, it might be enough to wrap "icon + text content" into an additional HStack with a `.xSmall(8)` spacing and leave the outer padding to `.medium` as it is.

Sources/Orbit/Components/ListChoice.swift
```diff
@@ -0,0 +1,252 @@
...
+    @ViewBuilder var trailingElementView: some View {
+        switch trailingElement {
+            case .none:
+                EmptyView()
+            case .disclosure:
+                Icon(symbol: .chevronRight, size: .medium, color: .inkNormal)
+            case .button(let type, let action):
+                switch type {
+                    case .add:
+                        Button(.plus, style: .primarySubtle, size: .small, action: action)
```
PavelHolec, line 78: Buttons will need their own vertical .small padding (12) to increase the height of the whole component. As this will make the switch cases more duplicated, it could be extracted to `@ViewBuilder func button(type:action:) -> some View`

Sources/Orbit/Components/ListChoice.swift
```diff
@@ -0,0 +1,252 @@
...
+    
+    @ViewBuilder var header: some View {
+        if isHeaderEmpty == false {
+            HStack(alignment: .firstTextBaseline, spacing: .small) {
+                iconView
+                    .alignmentGuide(.firstTextBaseline) { _ in
+                        Heading.Style.title3.size * 1.1
+                    }
+
+                HStack(alignment: .firstTextBaseline, spacing: .xxSmall) {
```
PavelHolec, line 55: This HStack seems to be "unused" (or at least the spacing part).

Sources/Orbit/Components/ListChoice.swift
```diff
@@ -0,0 +1,252 @@
...
+                Icon(symbol: .chevronRight, size: .medium, color: .inkNormal)
+            case .button(let type, let action):
+                switch type {
+                    case .add:
+                        Button(.plus, style: .primarySubtle, size: .small, action: action)
+                    case .remove:
+                        Button(.close, style: .criticalSubtle, size: .small, action: action)
+                }
+            case .checkbox(let isChecked, let isEnabled, let action):
+                Checkbox(state: isEnabled ? .normal : .disabled, isChecked: isChecked, action: action)
```
PavelHolec, line 83: The Checkbox here will be used as indicator only, meaning it should be `.disabled(true)`.
Similar with the `Button`s (it is not recommended to nest Buttons inside each other, it might not even work). 

The whole ListChoice component will probably have just a single `action` and the trailing elements will not be interactive on their own. (Similar behavior as in ChoiceTile, where radio/checkbox serve only as an non-interactive indicator).

Sources/Orbit/Components/ListChoice.swift
sjavora: Hmm I was sure that I checked the title should be `regular`, but you're right. I'll take the easier way out and just add the `weight` for now.

Sources/Orbit/Components/ListChoice.swift
```diff
@@ -0,0 +1,266 @@
...
+    ) {
+        self.title = title
+        self.description = description
+        self.icon = .icon(icon, size: .default, color: .inkNormal)
+        self.trailingElement = trailingElement
+        self.backgroundColor = backgroundColor
+        self.action = action
+    }
+    
+    init(
```
PavelHolec, line 127: This init comment can mention the custom icon content (can be copied from Tile, Card etc.)

Sources/Orbit/Components/ListChoiceGroup.swift
```diff
@@ -0,0 +1,104 @@
...
+        Card(spacing: 0, padding: 0, style: .iOS, status: status, backgroundColor: backgroundColor) {
+            content()
+        }
+    }
+}
+
+// MARK: - Inits
+public extension ListChoiceGroup {
+    
+    init(
```
PavelHolec, line 28: An init comment can be copied from Card.

Sources/Orbit/Components/ListChoice.swift
```diff
@@ -0,0 +1,266 @@
...
+
+            trailingElementView
+                .padding(.trailing, .medium)
+        }
+        .frame(maxWidth: Layout.readableMaxWidth, alignment: .leading)
+        .overlay(separator, alignment: .bottom)
+    }
+    
+    @ViewBuilder var header: some View {
+        if isHeaderEmpty == false {
```
PavelHolec, line 48: This check can be removed, as the Header (as other similar components like Text) already internally check whether they have "anything to show at all" and, if not, return EmptyView.

PavelHolec:
> Personally, I see little value in providing single-line comments to `init`s that say a type will be created... that's what `init`s do 😄
> 
> If we want to be comprehensive, we should document all public declarations, including function/init arguments.

Agree. For some components, there is more useful comment. But the most important part of that comment is the "... "Orbit component"..." which helps caller to be sure this is indeed an Orbit component, not a SwiftUI one (as we shadow some existing components with the same name)

PavelHolec (approved):

PR #6, opened 2022-01-27
Release action

sjavora:
We want a single-click workflow that will:

- [x] run stuff from the Automation folder
- [x] push changes
- [x] tag the latest commit on `main`
- [x] create a release with the provided version name and changelog
- [ ] update status of components in https://github.com/kiwicom/orbit/blob/master/docs/src/data/component-status.yaml either by pushing directly or by creating a PR in that repository

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -2,19 +2,35 @@ name: Release
 
 on:
   workflow_dispatch:
-    # Inputs the workflow accepts.
     inputs:
       version:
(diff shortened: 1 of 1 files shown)


sjavora:
Looks like this has become redundant. The automation scripts are now run on schedule, draft release is created by release-drafter and tagging is done when the draft is released.

Finally, the component status update will not be solved from this repository, but probably from the other end.

PR #16, opened 2022-02-02
Added HorizontalScroll component.

PavelHolec:
image

image

image

image


Sources/Orbit/Components/HorizontalScroll.swift
```diff
@@ -0,0 +1,235 @@
+import SwiftUI
+
+public enum HorizontalScrollItemWidth {
+    /// Each item will use its own intrinsic width.
+    case intrinsic
+    /// Width ratio compared to container width. Default value is `0.8`.
```
sjavora, line 6: > Default value is `0.8`.
Isn't this visible in quick help these days? I'd avoid restating it here because the comment can go out of sync with the implementation.

Sources/Orbit/Components/HorizontalScroll.swift
```diff
@@ -0,0 +1,235 @@
...
+            ScrollView(.horizontal, showsIndicators: false) {
+                HStack(alignment: .top, spacing: spacing) {
+                    content()
+                        .frame(width: itemWidth(forContentWidth: width), alignment: .leading)
+                        .frame(maxWidth: maxItemWidth, alignment: .leading)
+                        .frame(height: resolvedItemHeight, alignment: .top)
+                    
+                    // Provide reliable minimal height
+                    Color.clear
+                        .frame(width: 0, height: minHeight)
```
sjavora, line 49: This is at least the third time I see this technique - worth a utility view?

PavelHolec: Good idea, added a `Strut` component for this purpose.

Sources/Orbit/Components/HorizontalScroll.swift
PavelHolec: ✔️

sjavora (approved):

PR #19, opened 2022-02-04
Input field suffix

PavelHolec:
Add interactive suffix action to InputField

image

image

### DIFF
--- a/Sources/Orbit/Components/InputField.swift
+++ b/Sources/Orbit/Components/InputField.swift
@@ -14,11 +14,12 @@ public struct InputField: View {
 
     @Binding private var value: String
     @Binding private var messageHeight: CGFloat
-    @State private var isPressed: Bool = false
     @State private var isEditing: Bool = false
 
     let label: String
     let placeholder: String
+    let prefix: Icon.Content
+    let suffix: Icon.Content
     let state: InputState
     let textContent: UITextContentType?
     let keyboard: UIKeyboardType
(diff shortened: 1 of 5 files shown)


Sources/Orbit/Components/InputField.swift
```diff
@@ -14,11 +14,12 @@ public struct InputField: View {
 
     @Binding private var value: String
     @Binding private var messageHeight: CGFloat
-    @State private var isPressed: Bool = false
     @State private var isEditing: Bool = false
 
     let label: String
     let placeholder: String
+    let prefix: Icon.Content
+    let suffix: Icon.Content
```
sjavora, line 22: What is the logic in naming this one suffix, but calling it disclosure in other components? If it's "because Orbit calls it that", then I suggest making them unify it, because, this is just confusing.

Sources/Orbit/Support/Layout/Strut.swift
```diff
@@ -5,7 +5,7 @@ import SwiftUI
 /// Used to fix layout issues where `frame(minHeight:)` causes issues.
 struct Strut: View {
     
-    let height: CGFloat
+    let height: CGFloat?
```
sjavora, line 8: What is this change related to?

sjavora:
See, this is a reason you don't name elements "suffix" 😄 

Screenshot 2022-02-04 at 10 38 38

Sources/Orbit/Support/Layout/Strut.swift
sjavora: Oh I see... the module currently doesn't build. I wonder how we got into this situation given that each pull request builds the package...

Sources/Orbit/Components/InputField.swift
PavelHolec: Yep, so far it is the inconsistency in Orbit docs :/ I would address it later. For now, it seemed better if it matches the relevant component naming (although the disclosure is yet another term which should be renamed) 

image

image

Sources/Orbit/Support/Layout/Strut.swift
PavelHolec: 🤷

PavelHolec:
> See, this is a reason you don't name elements "suffix" 😄
> 
> Screenshot 2022-02-04 at 10 38 38

We need to tell Orbit then :D

sjavora:
@PavelHolec well all my comments have to do with naming... AFAIK you are the go-to person when it comes to Orbit <-> Mobile/iOS communication, so please bring the inconsistencies up 🙏

sjavora (approved):

PR #20, opened 2022-02-04
ListChoiceGroupStyle

sjavora:
There are two styles - `card`, which does what `ListChoiceGroup` did previously by default and `borderless`, which is also stripped down from `Card`, but has no borders/corner radius and always has white background.

Also added an optional `title` to `ListChoiceGroup`.

Screenshot 2022-02-04 at 12 52 46

### DIFF
--- a/Sources/Orbit/Components/ListChoiceGroup.swift
+++ b/Sources/Orbit/Components/ListChoiceGroup.swift
@@ -1,5 +1,14 @@
 import SwiftUI
 
+public enum ListChoiceGroupStyle {
+    
+    // Style with no decoration.
+    case borderless
+    
+    // Style with Card-like appearance.
+    case card(status: Status? = nil, backgroundColor: Color = .white)
+}
+
 /// Wraps ListChoice items.
 
 /// - Related components:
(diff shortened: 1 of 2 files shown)


sjavora:
Hmmm it seems like the build action passes even though the build actually failed 😞

Sources/Orbit/Components/ListChoiceGroup.swift
```diff
@@ -1,5 +1,14 @@
 import SwiftUI
 
+public enum ListChoiceGroupStyle {
+    
+    // Style with no decoration.
+    case borderless
+    
+    // Style with Card-like appearance.
+    case card(status: Status? = nil, backgroundColor: Color? = .clear)
```
PavelHolec, line 9: Default value here should be `nil` in case of optionals - if we further use it to distinguish "no change, let caller decide later". I believe `nil` it will later default to `.white` which is probably what we want in the end?

Sources/Orbit/Components/ListChoiceGroup.swift
```diff
@@ -11,28 +20,79 @@ public struct ListChoiceGroup<Content: View>: View {
...
+                }
+            case .borderless:
+                VStack(alignment: .leading, spacing: .medium) {
+                    header
+
+                    VStack(alignment: .leading, spacing: 0) {
+                        content()
+                    }
+                }
+                .background(Color.white)
```
PavelHolec, line 43: Borderless variant should be with no background.

Sources/Orbit/Components/ListChoiceGroup.swift
```diff
@@ -11,28 +20,79 @@ public struct ListChoiceGroup<Content: View>: View {
...
+                .frame(maxWidth: Layout.readableMaxWidth, alignment: .leading)
+                .frame(maxWidth: .infinity)
+                .padding(.horizontal, horizontalPadding)
+        }
+    }
+
+    @ViewBuilder var header: some View {
+        if isHeaderEmpty == false {
+            Header(title, iconContent: iconContent, titleStyle: titleStyle)
```
PavelHolec, line 52: `Header` should be able to decide emptiness on its own now (and return EmptyView in that case) - if not, we should fix it to avoid wrapping Orbit components with ifs.

PavelHolec:
As for now, design requests to have a smaller spacing between heading and the rest. It means it will be different from the Card.

Sources/Orbit/Components/ListChoiceGroup.swift
sjavora: `Card` has
```
backgroundColor: Color? = .white
```
in all its initializers - should we change those too?

Sources/Orbit/Components/ListChoiceGroup.swift
sjavora: That makes previews look funny, since `ListChoice` views have a default white background. What do we do about that?

PavelHolec (approved):

PR #21, opened 2022-02-04
Add `value` text view to `ListChoice`

sjavora:
Adds `value` that we talked about. When looking at it again, I finally understood what you meant with the extra stacks. `value` works with all "disclosure"s and even without them. The position of the custom content should now be the same as before. Notably, `value` is centered - the first baseline does not look good next to the chevron, or anything else really.

Screenshot 2022-02-04 at 22 31 46
Screenshot 2022-02-04 at 22 31 53
Screenshot 2022-02-04 at 22 31 59

### DIFF
--- a/Sources/Orbit/Components/ListChoice.swift
+++ b/Sources/Orbit/Components/ListChoice.swift
@@ -25,6 +25,7 @@ public struct ListChoice<Content: View>: View {
     let title: String
     let description: String
     let icon: Icon.Content
+    let value: String
     let disclosure: ListChoiceDisclosure
     let showSeparator: Bool
     let action: () -> Void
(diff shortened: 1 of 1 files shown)


PavelHolec:
Updated in new commit:
- Optimized layout, removed no longer used code
- Used `.heading4` for value as it better matches the header (until Orbit decides on final style)
- added few more previews

image

image

image

sjavora:
@PavelHolec well, I still can't approve this since I'm the author 🤷

PavelHolec (approved):

PavelHolec:
Added one more update for minimal vertical padding around header content

image

PR #28, opened 2022-02-11
Allow custom content in Alert.

PavelHolec:
image

### DIFF
--- a/Sources/Orbit/Components/Alert.swift
+++ b/Sources/Orbit/Components/Alert.swift
@@ -1,5 +1,17 @@
 import SwiftUI
 
+public enum AlertButtons {
+    case none
+    case primary(_ title: String, action: () -> Void = {})
+    case secondary(_ title: String, action: () -> Void = {})
+    case primaryAndSecondary(
+        primaryTitle: String,
+        secondaryTitle: String,
+        primaryAction: () -> Void = {},
+        secondaryAction: () -> Void = {}
+    )
+}
+
 /// Breaks the main user flow to present information.
 ///
 /// There are times when just simple information isn’t enough and the user needs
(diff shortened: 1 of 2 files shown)


Sources/Orbit/Components/Alert.swift
```diff
@@ -18,15 +30,16 @@ import SwiftUI
...
 /// - Note: [Orbit definition](https://orbit.kiwi/components/alert/)
 /// - Important: Component expands horizontally to infinity.
-public struct Alert: View {
+public struct Alert<Content: View>: View {
 
     let title: String
     let description: String
     let icon: Icon.Symbol
-    let buttons: Buttons
+    let buttons: AlertAction
```
sjavora, line 38: Shouldn't this be `action: AlertAction` now?

Sources/Orbit/Components/Alert.swift
```diff
@@ -42,13 +55,18 @@ public struct Alert: View {
...
+                
+                switch buttons {
+                    case .primary, .secondary, .primaryAndSecondary:
+                        // Keeping the identity of buttons for correct animations
+                        buttonsView
+                    case .none:
+                        EmptyView()
+                }
             }
+            .padding(.leading, icon == .none ? 0 : 28)
```
sjavora, line 69: Is there no constant for 28?

sjavora (changes requested):

Sources/Orbit/Components/Alert.swift
PavelHolec: Sadly no

Sources/Orbit/Components/Alert.swift
PavelHolec: Reverted the name to AlertButton.

sjavora (approved):

PR #38, opened 2022-02-20
Add script to update icons automatically

sjavora:
See https://github.com/kiwicom/orbit-swiftui/pull/39 for an example in action when triggered as part of tokens update.

This PR adds the changes manually, because some icons had custom names (`emptyCircle` -> `circleEmpty`) so references had to be adjusted.

Resolves https://github.com/kiwicom/orbit-swiftui/issues/29.


Sources/Orbit/Foundation/Icons/Icons.swift
```diff
@@ -177,243 +140,380 @@ public extension Icon {
...
-
-        case billingDetails
-        case customerSupport
-        case security
-
-        case attachment
-        case send
-
-        public var value: String {
+        var value: String {
```
sjavora, line 261: @PavelHolec  I made this `internal` - are we depending on this somewhere in the app?

PavelHolec (approved):

PR #46, opened 2022-03-01
Implement Secured InputField component

Parabak:
min iOS version is bumped to v14 to use `onChange(of:perform:)` 

I've added SecuredField as an option of InputField, but maybe it should be a completely separate component?

### DIFF
--- a/Sources/Orbit/Components/InputField.swift
+++ b/Sources/Orbit/Components/InputField.swift
@@ -90,15 +90,13 @@ public struct InputField: View {
             isEditing: $isEditing,
             style: .init(
                 textContentType: textContent,
-                autocapitalization: autocapitalization,
                 keyboardType: keyboard,
-                font: .orbit(size: Text.Size.normal.value, weight: .regular)
-            )
+                font: .orbit(size: Text.Size.normal.value, weight: .regular),
+                state: state
+            ),
+            onEditingChanged: onEditingChanged,
+            onCommit: onCommit
         )
-        .onTapGesture {
-            self.isEditing = true
-            onEditingChanged(isEditing)
-        }
         .background(textFieldPlaceholder, alignment: .leading)
     }
 
@@ -134,12 +132,14 @@ public struct InputField: View {
     }
 
     @ViewBuilder var securedSuffix: some View {
-        Icon(isSecureTextEntry ? .visibility : .visibilityOff, size: .default)
-            .padding(.horizontal, .xSmall)
-            .contentShape(Rectangle())
-            .onTapGesture {
-                isSecureTextEntry.toggle()
-            }
+        if value.isEmpty == false, state != .disabled {
+            Icon(isSecureTextEntry ? .visibility : .visibilityOff, size: .default)
+                .padding(.horizontal, .small)
+                .contentShape(Rectangle())
+                .onTapGesture {
+                    isSecureTextEntry.toggle()
+                }
+        }
     }
 }
 
--- a/Sources/Orbit/Components/InputField.swift
+++ b/Sources/Orbit/Components/InputField.swift
@@ -1,4 +1,5 @@
 import SwiftUI
+import UIKit
 
 /// Also known as textbox. Offers users a simple input for a form.
 ///
@@ -15,6 +16,7 @@ public struct InputField: View {
     @Binding private var value: String
     @Binding private var messageHeight: CGFloat
     @State private var isEditing: Bool = false
+    @State private var isSecureTextEntry: Bool = true
 
     let label: String
     let placeholder: String
@@ -25,8 +27,8 @@ public struct InputField: View {
     let keyboard: UIKeyboardType
     let autocapitalization: UITextAutocapitalizationType
     let isAutocompleteEnabled: Bool
+    let isSecure: Bool
     let message: MessageType
-
     let onEditingChanged: (Bool) -> Void
     let onCommit: () -> Void
     let suffixAction: () -> Void
@@ -34,19 +36,34 @@ public struct InputField: View {
     public var body: some View {
         VStack(alignment: .leading, spacing: .xxSmall) {
             FormFieldLabel(label)
-            
+
             InputContent(
                 prefix: prefix,
                 suffix: suffix,
                 state: state,
                 message: message,
                 isEditing: isEditing,
-                suffixAction: suffixAction
-            ) {
-                HStack(spacing: 0) {
-                    textField
-                    Spacer(minLength: 0)
-                    clearButton
+                suffixAction: suffixAction,
+                label: {
+                    HStack(spacing: 0) {
+                        input
+                            .textFieldStyle(TextFieldStyle(leadingPadding: 0))
+                            .autocapitalization(autocapitalization)
+                            .disableAutocorrection(isAutocompleteEnabled == false)
+                            .textContentType(textContent)
+                            .keyboardType(keyboard)
+                            .font(.orbit(size: Text.Size.normal.value, weight: .regular))
+                            .accentColor(.blueNormal)
+                            .frame(height: Layout.preferredButtonHeight)
+                            .background(textFieldPlaceholder, alignment: .leading)
+                            .disabled(state == .disabled)
+                        if isSecure {
+                            securedSuffix
+                        } else {
+                            Spacer(minLength: 0)
+                            clearButton
+                        }
+                    }
                 }
             }
             
@@ -56,6 +73,23 @@ public struct InputField: View {
                 }
             }
         }
+    }
+
+    @ViewBuilder var input: some View {
+        if isSecure {
+            secureField
+        } else {
+            textField
+        }
+    }
+
+    @ViewBuilder var secureField: some View {
+        SecureTextField(text: $value, isSecured: $isSecureTextEntry)
+        .onTapGesture {
+            self.isEditing = isEditing
+            onEditingChanged(isEditing)
+        }
+        .background(textFieldPlaceholder, alignment: .leading)
     }
 
     @ViewBuilder var textField: some View {
@@ -68,16 +102,6 @@ public struct InputField: View {
             },
             onCommit: onCommit
         )
-        .textFieldStyle(TextFieldStyle(leadingPadding: 0))
-        .autocapitalization(autocapitalization)
-        .disableAutocorrection(isAutocompleteEnabled == false)
-        .textContentType(textContent)
-        .keyboardType(keyboard)
-        .font(.orbit(size: Text.Size.normal.value, weight: .regular))
-        .accentColor(.blueNormal)
-        .frame(height: Layout.preferredButtonHeight)
-        .background(textFieldPlaceholder, alignment: .leading)
-        .disabled(state == .disabled)
     }
 
     @ViewBuilder var textFieldPlaceholder: some View {
@@ -96,6 +120,15 @@ public struct InputField: View {
                 .onTapGesture {
                     value = ""
                 }
+        }
+    }
+
+    @ViewBuilder var securedSuffix: some View {
+        Icon.Content.icon(isSecureTextEntry ? .visibility : .visibilityOff, size: .default)
+        .view()
+        .padding(.horizontal, .xSmall)
+        .onTapGesture {
+            isSecureTextEntry.toggle()
         }
     }
 }
@@ -119,6 +152,7 @@ public extension InputField {
         keyboard: UIKeyboardType = .default,
         autocapitalization: UITextAutocapitalizationType = .none,
         isAutocompleteEnabled: Bool = false,
+        isSecure: Bool = false,
         message: MessageType = .none,
         messageHeight: Binding<CGFloat> = .constant(0),
         onEditingChanged: @escaping (Bool) -> Void = { _ in },
@@ -137,6 +171,7 @@ public extension InputField {
         self.keyboard = keyboard
         self.autocapitalization = autocapitalization
         self.isAutocompleteEnabled = isAutocompleteEnabled
+        self.isSecure = isSecure
         self.onEditingChanged = onEditingChanged
         self.onCommit = onCommit
         self.suffixAction = suffixAction
@@ -162,15 +197,64 @@ public extension InputField {
     }
 }
 
+private struct SecureTextField: UIViewRepresentable {
+    typealias UIViewType = UITextField
+
+    @Binding var text: String
+    @Binding var isSecured: Bool
+
+    func makeUIView(context: Context) -> UITextField {
+        let textFied = UITextField()
+        textFied.text = text
+        textFied.isSecureTextEntry = isSecured
+
+        textFied.delegate = context.coordinator
+
+        return textFied
+    }
+
+    func updateUIView(_ uiView: UITextField, context: Context) {
+        uiView.isSecureTextEntry = isSecured
+
+        if uiView.isSecureTextEntry, let text = uiView.text {
+            uiView.text?.removeAll()
+            uiView.insertText(text)
+        }
+    }
+
+    func makeCoordinator() -> Coordinator {
+        Coordinator($text)
+    }
+
+    class Coordinator: NSObject, UITextFieldDelegate {
+        var text: Binding<String>
+
+        init(_ text: Binding<String>) {
+            self.text = text
+        }
+
+        public func textFieldDidEndEditing(_ textField: UITextField) {
+            self.text.wrappedValue = textField.text ?? ""
+        }
+
+        public func textFieldShouldReturn(_ textField: UITextField) -> Bool {
+            self.text.wrappedValue = textField.text ?? ""
+            return true
+        }
+    }
+}
+
 // MARK: - Previews
 struct InputFieldPreviews: PreviewProvider {
 
     static var previews: some View {
-        PreviewWrapper {
-            standalone
-            snapshots
+        Group {
+            PreviewWrapper {
+                standalone
+                snapshots
+            }
+            .previewLayout(PreviewLayout.sizeThatFits)
         }
-        .previewLayout(PreviewLayout.sizeThatFits)
     }
 
     static var standalone: some View {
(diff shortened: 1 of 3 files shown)


Package.swift
```diff
@@ -3,7 +3,7 @@ import PackageDescription
 
 let package = Package(
     name: "Orbit",
-    platforms: [.iOS(.v13)],
+    platforms: [.iOS(.v14)],
```
PavelHolec, line 6: We don't want to raise the iOS minimal version for all clients just because of the secure version of this component. We will probably raise it when absolutely necessary in future.

Let's only add an `if available` branch and add a new additional init so that the init for `isSecure` is available only for >=14.

Sources/Orbit/Components/InputField.swift
```diff
@@ -160,15 +215,45 @@ public extension InputField {
     }
 }
 
+// MARK: - Model
+private class SecureViewModel: ObservableObject {
```
PavelHolec, line 219: Do we need a model + combine for that? 

If we do, it would need to be a `@StateObject` though as @ObervedObject lifetime cannot be handled by the same view itself.

Do we also need to run it every 3 seconds? 

Can we instead use 
```
Task.sleep()
isShowSecuredText
```

directly in the suffix action?

Sources/Orbit/Components/InputField.swift
```diff
@@ -160,15 +215,45 @@ public extension InputField {
...
-        PreviewWrapper {
-            standalone
-            snapshots
+        Group {
+            PreviewWrapper {
+                standalone
+                snapshots
+            }
+            .previewLayout(PreviewLayout.sizeThatFits)
+            PreviewWrapper {
```
PavelHolec, line 251: I cant see a difference between these two grouped wrappers 🤔

Sources/Orbit/Components/InputField.swift
```diff
@@ -255,7 +342,34 @@ struct InputFieldLivePreviews: PreviewProvider {
...
             .animation(.easeOut(duration: 0.25), value: message)
             .padding()
-            .previewDisplayName("Run Live Preview")
+            .previewDisplayName("Run Live Preview with Input Field")
+        }
+    }
+
+
+    struct PreviewWrapperSecured: View {
+        @State var message: MessageType = .none
```
PavelHolec, line 351: Does the message change at all? It seems to be not bound to anything. A `PreviewWrapperWithState` could be used instead to handle the `value`.

Sources/Orbit/Components/InputField.swift
```diff
@@ -37,25 +39,58 @@ public struct InputField: View {
...
+        SecureField(
+            "",
+            text: $value
+        ).onTapGesture {
+            self.isEditing = isEditing
+            onEditingChanged(isEditing)
+        }
+        .textFieldStyle(TextFieldStyle(leadingPadding: 0))
+        .autocapitalization(autocapitalization)
+        .disableAutocorrection(isAutocompleteEnabled == false)
```
PavelHolec, line 81: Probably most of these modifiers are the same as for non-secure version. It should be extracted to vary only what is truly different between those two.

PavelHolec (changes requested):

Sources/Orbit/Components/InputField.swift
Parabak: > Do we need a model + combine for that?
I fully understand your concern but I failed to do it without the ViewModel. View itself can't change its state, but `isShowSecuredText` has to be switched back to `false`  after some delay (e.g. 3 seconds) to show the secured input

>Do we also need to run it every 3 seconds?
We don't, here we just wait for one event with the help of `.first()` operator and then the publisher finishes.

Task can be used here as well. I've tried it at first, but compiler was complaining about modifying captured self inside. However it shouldn't be the case if the model is a `class`. So I can update it with `Task`

Sources/Orbit/Components/InputField.swift
Parabak: good catch, it's a duplication. Will fix

Sources/Orbit/Components/InputField.swift
Parabak: What is the way to achieve? Should it be something like
```
if isSecured {
   SecureField()
   .customeModifier1
} else {
   TextField
   .customeModifier2
}
.commonModifier1
.commonModifier2
```

Sources/Orbit/Components/InputField.swift
PavelHolec: I mean - we might not even need to use the Model in the first place (possibly)

Sources/Orbit/Components/InputField.swift
PavelHolec: Yes exactly 👍

If it gets too long, it can be further wrapped as 

```
var body: some View {
   if isSecured {
      secureField
   } else {
      textField
}

var secureField: some View {
   inputField
      .customMod2
}

var textField: some View {
   inputField
      .customMod1
}

var inputField: some View {
   //common code
}
```

Sources/Orbit/Components/InputField.swift
```diff
@@ -96,6 +131,22 @@ public struct InputField: View {
...
+                secureModel.isShowSecuredText = false
+            }
+            .opacity(secureModel.isShowSecuredText ? 1 : 0)
+            .animation(.easeOut(duration: 0.2), value: secureModel.isShowSecuredText)
+    }
+
+    var securedSuffix: Icon.Content {
+        secureModel.isShowSecuredText
+            ? Icon.Content.icon(.visibilityOff, size: .default)
+            : Icon.Content.icon(.visibility, size: .default)
```
sjavora, line 148: `Icon.Content.icon(secureModel.isShowSecuredText ? .visibilityOff : .visibility, size: .default)`?

Sources/Orbit/Components/InputField.swift
```diff
@@ -160,15 +215,45 @@ public extension InputField {
     }
 }
 
+// MARK: - Model
+private class SecureViewModel: ObservableObject {
+
+    @Published var isShowSecuredText = false {
```
sjavora, line 221: "isShow"?

As I don't know exactly what is meant here, some options that come to mind are `isShowingSecuredText`, `showSecureText`, `showsSecureText`...

Sources/Orbit/Components/InputField.swift
```diff
@@ -37,25 +39,58 @@ public struct InputField: View {
...
                 message: message,
                 isEditing: isEditing,
-                suffixAction: suffixAction
-            ) {
-                HStack(spacing: 0) {
-                    textField
-                    Spacer(minLength: 0)
-                    clearButton
+                suffixAction: {
+                    if isSecure {
```
PavelHolec, line 47: The "eye" icon suffix will be specific to the InputField and should not clash with any existing suffix. It needs its own icon and action.

The full possible order of suffixes will be:
"X(clear value) - Eye(toggle password) - Suffix(any caller suffix)"

We may or may not want to show X(clear) button on secure fields - normally we do, so we do not have to change it for now.

Sources/Orbit/Components/InputField.swift
Parabak: We need something that will update @State after some delay. I failed to create such mechanism inside of the View because View is not mutable — I can't assign Task, Cancellation later on...

Sources/Orbit/Components/InputField.swift
Parabak: So that means I have to split this logic: eye is a standalone element, right?

Sources/Orbit/Components/InputField.swift
Parabak: I've decided to extract field into separate var:
```
    @ViewBuilder var input: some View {
        if isSecure {
            secureField
        } else {
            textField
        }
    }
```

thought can't come up with some meaningful name...

Anyway, compiler doesn't allow me to use ternary operator saying 
> Result values in '? :' expression have mismatching types 'some View' (type of 'InputField.secureField') and 'some View' (type of 'InputField.textField')

for `isSecure ? secureField : textField`


Do you know why?

Parabak:
Hey guys, the MR is updated. Please have a look. I still need to test in an App, so far I've tried it in Preview

Sources/Orbit/Components/InputField.swift
sjavora: This is a function builder, so the whole thing desugars into method calls. Perhaps ternary operator support is not implemented.

Sources/Orbit/Components/InputField.swift
PavelHolec: Yep, the eye icon would be an additional view that will be added (to the HStack of existing suffixes) only when `isSecture = true`.

Sources/Orbit/Components/InputField.swift
```diff
@@ -96,6 +125,25 @@ public struct InputField: View {
...
+            .disabled(true)
+            .customOnChange(of: value, perform: { _ in
+                secureModel.showSecureText = false
+            })
+            .opacity(secureModel.showSecureText ? 1 : 0)
+            .animation(.easeOut(duration: 0.2), value: secureModel.showSecureText)
+    }
+
+    @ViewBuilder var securedSuffix: some View {
+        Icon.Content.icon(secureModel.showSecureText ? .visibilityOff : .visibility, size: .default)
```
PavelHolec, line 140: can be just `Icon(secureModel.showSecureText ? .visibilityOff : .visibility, size: .default)`

Parabak:
@PavelHolec , @sjavora updated with `UIViewRepresentable` and `UITextField`. Less custom extensions, animations, Combine — so please have a look, try in Preview and comment 😅

Sources/Orbit/Components/InputField.swift
```diff
@@ -160,15 +195,64 @@ public extension InputField {
     }
 }
 
+private struct SecureTextField: UIViewRepresentable {
```
PavelHolec, line 198: Ideally this should be moved to `Support` folder, along with its own preview (similar to existing UIViewRepresentable wrappers that are there)

Sources/Orbit/Components/InputField.swift
```diff
@@ -96,6 +120,15 @@ public struct InputField: View {
                 }
         }
     }
+
+    @ViewBuilder var securedSuffix: some View {
+        Icon.Content.icon(isSecureTextEntry ? .visibility : .visibilityOff, size: .default)
```
PavelHolec, line 125: Just `Icon(isSecureTextEntry ? .visibility : .visibilityOff)`

Sources/Orbit/Components/InputField.swift
```diff
@@ -160,15 +195,64 @@ public extension InputField {
...
+}
+
 // MARK: - Previews
 struct InputFieldPreviews: PreviewProvider {
 
     static var previews: some View {
-        PreviewWrapper {
-            standalone
-            snapshots
+        Group {
```
PavelHolec, line 249: Can be reverted.

Sources/Orbit/Components/InputField.swift
```diff
@@ -160,15 +195,64 @@ public extension InputField {
...
 }
 
+private struct SecureTextField: UIViewRepresentable {
+    typealias UIViewType = UITextField
+
+    @Binding var text: String
+    @Binding var isSecured: Bool
+
+    func makeUIView(context: Context) -> UITextField {
+        let textFied = UITextField()
```
PavelHolec, line 205: The textField is not using Orbit font. Size, font and color should match the normal input field.

PavelHolec:
Binding seems to be ignored:

https://user-images.githubusercontent.com/35844477/158986840-8661b27c-3e06-443a-a806-a71663f57e75.MP4

PavelHolec:
Icon color and alignment do not match InputField

image

Sources/Orbit/Components/InputField.swift
```diff
@@ -96,6 +120,15 @@ public struct InputField: View {
                 }
         }
     }
+
+    @ViewBuilder var securedSuffix: some View {
+        Icon.Content.icon(isSecureTextEntry ? .visibility : .visibilityOff, size: .default)
+        .view()
+        .padding(.horizontal, .xSmall)
+        .onTapGesture {
```
PavelHolec, line 128: It needs `.contentShape(Rectangle())` before this line as now it is hard to tap.

Parabak: @PavelHolec
> Binding seems to be ignored:

It's updated when editing ends... I can implement handling each symbol, but is it required? It will require some handling in `shouldChangeCharactersIn` when user deletes char or paste something...

PavelHolec: We need both, just like the regular InputField (api needs to match):
1) current value (“typing in progress”) binding (to catch “password to short” and similar validation in realtime)
2) and the regular “typing is done”

Sources/Orbit/Components/InputField.swift
Parabak: @PavelHolec updated. Please review all opened threads, they should be done as well 🙏

Sources/Orbit/Components/InputField.swift
```diff
@@ -25,37 +27,79 @@ public struct InputField: View {
...
+                            .textFieldStyle(TextFieldStyle(leadingPadding: 0))
+                            .autocapitalization(autocapitalization)
+                            .disableAutocorrection(isAutocompleteEnabled == false)
+                            .textContentType(textContent)
+                            .keyboardType(keyboard)
+                            .font(.orbit(size: Text.Size.normal.value, weight: .regular))
+                            .accentColor(.blueNormal)
+                            .frame(height: Layout.preferredButtonHeight)
+                            .background(textFieldPlaceholder, alignment: .leading)
+                            .disabled(state == .disabled)
```
PavelHolec, line 59: Secure version does not respect disabled color style (looks as if it is active).

Sources/Orbit/Components/InputField.swift
```diff
@@ -15,6 +16,7 @@ public struct InputField: View {
     @Binding private var value: String
     @Binding private var messageHeight: CGFloat
     @State private var isEditing: Bool = false
+    @State private var isSecureTextEntry: Bool = true
 
     let label: String
     let placeholder: String
```
PavelHolec, line 22: The placeholder is displayed below the password value when the field is disabled.

Sources/Orbit/Components/InputField.swift
```diff
@@ -25,37 +27,79 @@ public struct InputField: View {
     let keyboard: UIKeyboardType
     let autocapitalization: UITextAutocapitalizationType
     let isAutocompleteEnabled: Bool
+    let isSecure: Bool
     let message: MessageType
-
     let onEditingChanged: (Bool) -> Void
     let onCommit: () -> Void
```
PavelHolec, line 33: Secure version does not call `onCommit`

Sources/Orbit/Support/Views/SecureTextField.swift
```diff
@@ -0,0 +1,87 @@
+import SwiftUI
+import UIKit
+
+struct SecureTextFieldStyle {
+    let textContentType: UITextContentType?
+    let autocapitalization: UITextAutocapitalizationType
```
PavelHolec, line 6: `autocapitalization` does not seem to be respected in secure version. (We can remove it if it is hard to implement)

PavelHolec (changes requested):

Sources/Orbit/Support/Views/SecureTextField.swift
```diff
@@ -0,0 +1,87 @@
...
+
+        public func textFieldShouldReturn(_ textField: UITextField) -> Bool {
+            endEditing(textField)
+            return true
+        }
+
+        func textField(_ textField: UITextField, shouldChangeCharactersIn range: NSRange, replacementString string: String) -> Bool {
+
+            if let input = textField.text, let specialRange = Range(range, in: input) {
+                text.wrappedValue.replaceSubrange(specialRange, with: string)
```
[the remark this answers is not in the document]
Parabak: > (simulated by clearing other field with the same binding)

Do you mean by clearing the same binded value in nonsecured text field?

Sources/Orbit/Support/Views/SecureTextField.swift
```diff
@@ -0,0 +1,87 @@
...
+            textFied.becomeFirstResponder()
+        }
+
+        return textFied
+    }
+
+    func updateUIView(_ uiView: UITextField, context: Context) {
+        uiView.isSecureTextEntry = isSecured
+
+        if uiView.isSecureTextEntry && uiView.text == text {
```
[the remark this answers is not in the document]
Parabak: Could you please elaborate — in what case binding is broken?

Sources/Orbit/Support/Views/SecureTextField.swift
```diff
@@ -0,0 +1,87 @@
...
+            // Workaround. Without it, UITextField will erase it's own current value on frist input
+            uiView.text?.removeAll()
```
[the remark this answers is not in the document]
Parabak: Can't reproduce it... Maybe you can comment it out and just verify is it that reason or it's somewhere else?

Sources/Orbit/Support/Views/SecureTextField.swift
Parabak: I was thinking does it make sense at all in password context. Let's remove it for now

Sources/Orbit/Support/Views/SecureTextField.swift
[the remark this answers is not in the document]
PavelHolec: It is subtle, so we can leave it for later:

https://user-images.githubusercontent.com/35844477/159518569-13690a6e-7d34-428f-b5f9-eb6cd0478849.mov

Sources/Orbit/Support/Views/SecureTextField.swift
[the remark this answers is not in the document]
PavelHolec: Use case:
https://user-images.githubusercontent.com/35844477/159518839-9150e609-7402-4546-9dd6-2cbecab47b11.mov

Sources/Orbit/Support/Views/SecureTextField.swift
PavelHolec: Sure, but the InputField still has this argument and it will be silently ignored 🤷  
We can fix it later, but let's keep FIXME: comment somewhere.

Sources/Orbit/Support/Views/SecureTextField.swift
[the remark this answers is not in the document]
Parabak: Now I can see, thanks. That happens only with our custom font, if you comment it out — it stops jumping. Also I've checked our current implementation in Kiwi.app and it also jumps there 😅

Parabak:
updated yet again

PavelHolec (approved):

PR #59, opened 2022-03-15
Rename Heading to Label and unify spacing

PavelHolec:
image


sjavora (approved):

sjavora:
@PavelHolec don't know about the name... `Label` doesn't evoke "has an icon and maybe some other text" in my head, but otherwise 👍

username0x0a:
Agree with @sjavora, `Label` doesn't really have a semantic meaning in relation to Document text levels (where this should exactly be a `Heading` analogy), it's rather a component which carries different levels and styles of text, be it paragraph, heading, quote, etc. 🤔

Taking the snapshots in mind, it would rather be a `Title` to complement the `Description`, but this really starts being complicated.

PR #60, opened 2022-03-15
Replace ListChoiceGroup with Card

PavelHolec:
image


Sources/Orbit/Support/Views/TileBorder.swift
```diff
@@ -1,5 +1,12 @@
 import SwiftUI
 
+public enum TileBorderStyle {
+    case none
+    case `default`
+    /// A card style that visually matches the iOS plain table section appearance.
```
sjavora, line 6: This mentions "card" - not sure if it makes sense without some context.

I guess it was there already...

PavelHolec: Fixed

sjavora (approved):

PR #62, opened 2022-03-16
Update icon sizes

PavelHolec:
There are new and updated icon sizes. Code has been refactored to unify all icon-related images to use the same Icon size enum. Added CountryFlag as one of Icon options.

image

image

image

image

image

image

image

### DIFF
--- a/Sources/Orbit/Components/CountryFlag.swift
+++ b/Sources/Orbit/Components/CountryFlag.swift
@@ -7,24 +7,32 @@ import SwiftUI
 ///
 /// - Note: [Orbit definition](https://orbit.kiwi/components/countryflag/)
 public struct CountryFlag: View {
-
+    
     let countryCode: String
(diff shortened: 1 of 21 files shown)


Sources/Orbit/Components/CountryFlag.swift
```diff
@@ -8,71 +8,58 @@ import SwiftUI
...
+        }
     }
 }
 
 // MARK: - Inits
 public extension CountryFlag {
 
     /// Creates Orbit CountryFlag component.
-    init(_ countryCode: String, height: Height = .default, border: Border = .default) {
+    init(_ countryCode: String, size: Icon.Size = .normal, border: Border = .default) {
```
sjavora, line 45: Not sure how long this has been here, but don't we want an enum for the countries?

username0x0a (approved):

PR #81, opened 2022-03-28
PasswordStrengthIndicator

Parabak:
Adding PasswordStrengthIndicator to WIP folder

### DIFF
--- a/Sources/Orbit/Components/InputField.swift
+++ b/Sources/Orbit/Components/InputField.swift
@@ -28,6 +28,7 @@ public struct InputField: View {
     let autocapitalization: UITextAutocapitalizationType
     let isAutocompleteEnabled: Bool
     let isSecure: Bool
+    let passwordStrength: PasswordStrengthIndicator.PasswordStrength
     let message: MessageType
     let onEditingChanged: (Bool) -> Void
     let onCommit: () -> Void
@@ -35,31 +36,36 @@ public struct InputField: View {
 
     public var body: some View {
         FormFieldWrapper(label, message: message, messageHeight: $messageHeight) {
-            InputContent(
-                prefix: prefix,
-                suffix: suffix,
-                state: state,
-                message: message,
-                isEditing: isEditing,
-                suffixAction: suffixAction
-            ) {
-                HStack(spacing: 0) {
-                    input
-                        .textFieldStyle(TextFieldStyle(leadingPadding: 0))
-                        .autocapitalization(autocapitalization)
-                        .disableAutocorrection(isAutocompleteEnabled == false)
-                        .textContentType(textContent)
-                        .keyboardType(keyboard)
-                        .font(.orbit(size: Text.Size.normal.value, weight: .regular))
-                        .accentColor(.blueNormal)
-                        .background(textFieldPlaceholder, alignment: .leading)
-                        .disabled(state == .disabled)
-                    if isSecure {
-                        securedSuffix
-                    } else {
-                        clearButton
+            VStack(alignment: .leading, spacing: .xxSmall) {
+                InputContent(
+                    prefix: prefix,
+                    suffix: suffix,
+                    state: state,
+                    message: message,
+                    isEditing: isEditing,
+                    suffixAction: suffixAction
+                ) {
+                    HStack(spacing: 0) {
+                        input
+                            .textFieldStyle(TextFieldStyle(leadingPadding: 0))
+                            .autocapitalization(autocapitalization)
+                            .disableAutocorrection(isAutocompleteEnabled == false)
+                            .textContentType(textContent)
+                            .keyboardType(keyboard)
+                            .font(.orbit(size: Text.Size.normal.value, weight: .regular))
+                            .accentColor(.blueNormal)
+                            .background(textFieldPlaceholder, alignment: .leading)
+                            .disabled(state == .disabled)
+                        if isSecure {
+                            securedSuffix
+                        } else {
+                            clearButton
+                        }
                     }
                 }
+
+                PasswordStrengthIndicator(passwordStrength: passwordStrength)
+                    .padding(.top, .xxSmall)
             }
         }
     }
@@ -155,6 +161,7 @@ public extension InputField {
         autocapitalization: UITextAutocapitalizationType = .none,
         isAutocompleteEnabled: Bool = false,
         isSecure: Bool = false,
+        passwordStrength: PasswordStrengthIndicator.PasswordStrength = .empty,
         message: MessageType = .none,
         messageHeight: Binding<CGFloat> = .constant(0),
         onEditingChanged: @escaping (Bool) -> Void = { _ in },
@@ -174,6 +181,7 @@ public extension InputField {
         self.autocapitalization = autocapitalization
         self.isAutocompleteEnabled = isAutocompleteEnabled
         self.isSecure = isSecure
+        self.passwordStrength = passwordStrength
         self.onEditingChanged = onEditingChanged
         self.onCommit = onCommit
         self.suffixAction = suffixAction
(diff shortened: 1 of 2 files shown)


Sources/Orbit/WIP/PasswordStrengthIndicator.swift
```diff
@@ -0,0 +1,107 @@
...
+                        .padding(.trailing, geometry.size.width * (1 - max(min(1, progress), 0)))
+                        .animation(.easeOut, value: progress)
+                },
+                alignment: .leading
+            )
+    }
+
+    var text: String {
+        switch passwordStrength {
+            case .empty:    return ""
```
sjavora, line 44: Please align this with the other returns.

Sources/Orbit/WIP/PasswordStrengthIndicator.swift
```diff
@@ -0,0 +1,107 @@
+import SwiftUI
+
+public enum PasswordStrength {
```
sjavora, line 3: Wouldn't it make sense to nest this inside the view?

sjavora:
@Parabak why is this in the WIP folder then? Is it only supposed to be used internally in actual components?

Parabak:
> Is it only supposed to be used internally in actual components?

Good question. The next step will be to use it as part of InputField. However, we might consider making it `public` itself to be used wherever it's needed 🤔 wdyt?

Sources/Orbit/WIP/PasswordStrengthIndicator.swift
PavelHolec: Agree. It is usually driven by whether the View is generic or not (even though rarely, this sometimes changes from non-generic->generic)

Sources/Orbit/WIP/PasswordStrengthIndicator.swift
```diff
@@ -0,0 +1,107 @@
...
+            }
+        }
+    }
+
+    var indicator: some View {
+        Capsule()
+            .frame(height: 2)
+            .foregroundColor(Color.cloudDark)
+            .overlay(
+                GeometryReader { geometry in
```
PavelHolec, line 32: Now that the design got more constant, we can get rid of geometry reader and calculations.

image

Code:

```
import SwiftUI
import Orbit

struct ProgressView: View {

    enum Progress {
        case empty
        case weak
        case medium
        case strong
    }

    let progress: Progress

    var body: some View {
        HStack(spacing: 0) {
            Capsule()
                .fill(Color.cloudNormal)
                .overlay(progressBar)
                .frame(height: .xxSmall)
                .frame(maxWidth: .infinity, alignment: .leading)

            Text(text, color: .custom(color))
                .frame(maxWidth: .infinity, alignment: .trailing)
        }
    }

    @ViewBuilder var progressBar: some View {
        HStack(spacing: 0) {
            Capsule()
                .fill(Color(color))
                .frame(maxWidth: .infinity)

            ForEach(0 ..< spacersCount, id: \.self) { _ in
                Spacer()
                    .frame(maxWidth: .infinity)
            }
        }
    }

    var spacersCount: Int {
        switch progress {
            case .empty:        return 4
            case .weak:         return 3
            case .medium:       return 1
            case .strong:       return 0
        }
    }

    var color: UIColor {
        switch progress {
            case .empty:        return .clear
            case .weak:         return .redNormal
            case .medium:       return .orangeNormal
            case .strong:       return .greenNormal
        }
    }

    var text: String {
        switch progress {
            case .empty:        return "-"
            case .weak:         return "weak"
            case .medium:       return "medium"
            case .strong:       return "strong"
        }
    }
}

struct ProgressViewPreviews: PreviewProvider {
    
        PreviewWrapper {
            ProgressView(progress: .empty)
            ProgressView(progress: .weak)
            ProgressView(progress: .medium)
            ProgressView(progress: .strong)
        }
        .previewLayout(.sizeThatFits)
    }
}

```

Sources/Orbit/WIP/PasswordStrengthIndicator.swift
Parabak: As `PasswordStrength` is a param of new `InputField.init` it will stay public and therefore outside of PasswordStrengthIndicator which is still internal

Sources/Orbit/WIP/PasswordStrengthIndicator.swift
Parabak: Applied suggestion with one change — the whole component is hidden if the value is empty, no need for "-" text in that case according to the design.

Parabak:
Oh, I have to move password indicator into `FormFieldWrapper`...

PavelHolec:
The indicator should be animated (baked-in, fixed animation), ideally in a same way as the existing `Progress` component.

Sources/Orbit/Support/Forms/FormFieldWrapper.swift
```diff
@@ -16,6 +17,9 @@ public struct FormFieldWrapper<Content: View>: View {
 
             content()
 
+            PasswordStrengthIndicator(passwordStrength: passwordStrength)
```
PavelHolec, line 20: The indicator should be provided as a part of `content` from the parent view, not used directly in this generic wrapper.

Sources/Orbit/WIP/PasswordStrengthIndicator.swift
PavelHolec: That, or the indicator component could be a `public` component as well so that client can use it to construct similar components.

Sources/Orbit/Support/Forms/FormFieldWrapper.swift
Parabak: I've moved it mimic the FormFieldMessage 🤷 Will move it back to `InputField`

Parabak:
@PavelHolec updated
@sjavora could you have another look?

Sources/Orbit/WIP/PasswordStrengthIndicator.swift
Parabak: Missed that comment, will update

Parabak:
ready for review

PavelHolec:
The indicator should be animated in the same way as the existing `Progress` component

Sources/Orbit/Support/Forms/FormFieldWrapper.swift
```diff
@@ -1,6 +1,6 @@
 import SwiftUI
 
-/// Orbit wrapper around form fields. Provides optional label and message.
+/// Orbit wrapper around form fields. Provides optional label, password strength indicator and message.
```
PavelHolec, line 3: Should be reverted

Sources/Orbit/Support/Components/PasswordStrengthIndicator.swift
```diff
@@ -0,0 +1,118 @@
...
+    case weak(title: String)
+    case medium(title: String)
+    case strong(title: String)
+}
+
+/// Orbit password strength indicator.
+public struct PasswordStrengthIndicator: View {
+
+    let passwordStrength: PasswordStrength
+    let textAccessibilityIdentifier: String
```
PavelHolec, line 14: Probably no need for that, as the caller can simply assign the ID through the only StaticText that is here.

Sources/Orbit/Support/Components/PasswordStrengthIndicator.swift
```diff
@@ -0,0 +1,118 @@
...
+        switch passwordStrength {
+            case .empty:    return .clear
+            case .weak:     return .redNormal
+            case .medium:   return .orangeNormal
+            case .strong:   return .greenNormal
+        }
+    }
+}
+
+extension PasswordStrengthIndicator {
```
PavelHolec, line 78: Will need to be `public extension`, as other component inits

Sources/Orbit/Components/InputField.swift
```diff
@@ -345,6 +365,18 @@ struct InputFieldLivePreviews: PreviewProvider {
         }
     }
 
+    static func validate(password: String) -> PasswordStrength {
+        switch password.count {
+            case 0:
+                return .empty
```
PavelHolec, line 371: Can be one-line switch cases

Sources/Orbit/Components/InputField.swift
```diff
@@ -28,40 +28,47 @@ public struct InputField: View {
...
+                            clearButton
+                        }
                     }
                 }
+
+                PasswordStrengthIndicator(passwordStrength: passwordStrength)
+                    .padding(.top, .xxSmall)
             }
         }
+        .animation(.easeInOut(duration: 0.2))
```
PavelHolec, line 71: Animation without `value:` will be deprecated. Should be `, value: passwordStrength`

PavelHolec: One more thing I realized. If we bake-in the animation here, it will be out of sync with the parent stacks animations.
We probably should not animate the whole InputField at all, so that caller can decide if and how to animate the stacks.

Comparison:

Current version (out of sync animation)

https://user-images.githubusercontent.com/35844477/170255937-61c23a75-ec25-4697-9393-f63331ed9db2.mov

Proposed version (without this animation modifier and with applied .animation() to the specific preview VStack only)

https://user-images.githubusercontent.com/35844477/170256072-102d3c06-1c11-4bda-987f-7476f73a290b.mov

PavelHolec (approved):

sjavora (approved):

PR #90, opened 2022-04-05
Update dialog component

PavelHolec:
Finalized version in Figma:
https://www.figma.com/file/YMfJkmUMV8xiY7fgeW6QAr/%5BOrbit%5D-Native-Components?node-id=3267%3A647

image

image

### DIFF
--- a/Sources/Orbit/Components/Button.swift
+++ b/Sources/Orbit/Components/Button.swift
@@ -283,6 +283,22 @@ extension Button {
             }
         }
     }
+
+    public struct Content: ExpressibleByStringLiteral {
+        let label: String
+        let accessibilityIdentifier: String
+        let action: () -> Void
+
+        public init(_ label: String, accessibilityIdentifier: String = "", action: @escaping () -> Void = {}) {
+            self.label = label
+            self.accessibilityIdentifier = accessibilityIdentifier
+            self.action = action
+        }
+
+        public init(stringLiteral value: String) {
+            self.init(value)
+        }
+    }
 }
 
 // MARK: - Previews
--- a/Sources/Orbit/Components/Dialog.swift
+++ b/Sources/Orbit/Components/Dialog.swift
@@ -6,91 +6,112 @@ import SwiftUI
 public struct Dialog: View {
 
     public static let shadowColor = Color.inkNormal
-    
+
     let illustration: Illustration.Image
     let title: String
     let description: String
-    let buttons: ButtonConfiguration
+    let style: Style
+    let buttonConfiguration: Buttons
 
     public var body: some View {
         VStack(alignment: .center, spacing: .medium) {
-            if illustration != .none {
-                Illustration(illustration, layout: .frame(maxHeight: 120))
-                    .padding(.top, .medium)
-            }
+            Illustration(illustration, layout: .frame(maxHeight: 120))
+                .padding(.top, .medium)
 
             VStack(alignment: .center, spacing: .xSmall) {
-                Heading(title, style: .title4, alignment: .center)
-
+                Heading(title, style: .title3, alignment: .center)
                 Text(description, color: .inkLight, alignment: .center)
             }
 
             VStack(alignment: .center, spacing: .xSmall) {
-                switch buttons {
-                    case .primary(let title, let style, let action):
-                        Button(title, style: style.primaryStyle, action: action)
-                    case .primaryAndSecondary(
-                        let primaryTitle, let secondaryTitle, let style, let primaryAction, let secondaryAction
-                    ):
-                        Button(primaryTitle, style: style.primaryStyle, action: primaryAction)
-                        Button(secondaryTitle, style: style.secondaryStyle, action: secondaryAction)
-                }
+                buttons
             }
         }
         .padding(.medium)
         .background(
             RoundedRectangle(cornerRadius: .small)
                 .fill(.white)
-                .shadow(color: Self.shadowColor.opacity(0.3), radius: .medium, x: 0, y: .medium / 2)
+                .shadow(color: Self.shadowColor.opacity(0.7), radius: .xxxLarge, x: 0, y: .xxxLarge / 2)
         )
-        .overlay(
-            RoundedRectangle(cornerRadius: .small)
-                .strokeBorder(Self.shadowColor.opacity(0.15), lineWidth: BorderWidth.thin)
-        )
+        .frame(maxWidth: Layout.readableMaxWidth / 2, maxHeight: .infinity)
+        .padding(.xLarge)
+        .background(Color.inkNormal.opacity(0.45))
+        .edgesIgnoringSafeArea(.all)
     }
-    
+
+    @ViewBuilder var buttons: some View {
+        switch buttonConfiguration {
+            case .primary(let primaryButton),
+                 .primaryAndSecondary(let primaryButton, _),
+                 .primarySecondaryAndTertiary(let primaryButton, _, _):
+                Button(primaryButton.label, style: style.buttonStyle, action: primaryButton.action)
+                    .accessibility(identifier: primaryButton.accessibilityIdentifier)
+        }
+
+        switch buttonConfiguration {
+            case .primary:
+                EmptyView()
+            case .primaryAndSecondary(_, let secondaryButton),
+                 .primarySecondaryAndTertiary(_, let secondaryButton, _):
+                ButtonLink(secondaryButton.label, style: style.buttonLinkStyle, size: .button, action: secondaryButton.action)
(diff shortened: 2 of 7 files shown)


sjavora:
You can't "deny" notifications in a custom dialog 😅 

I hope designers don't get the wrong idea...

PavelHolec:
> You can't "deny" notifications in a custom dialog 😅
> 
> I hope designers don't get the wrong idea...

Updated

Sources/Orbit/Components/Dialog.swift
```diff
@@ -6,91 +6,109 @@ import SwiftUI
...
-        )
-        .overlay(
-            RoundedRectangle(cornerRadius: .small)
-                .strokeBorder(Self.shadowColor.opacity(0.15), lineWidth: BorderWidth.thin)
+                .shadow(color: Self.shadowColor.opacity(0.7), radius: .xxxLarge, x: 0, y: .xxxLarge / 2)
         )
+        .frame(maxWidth: Layout.readableMaxWidth / 2, maxHeight: .infinity)
+        .padding(.xLarge)
+        .background(Color.inkNormal.opacity(0.45))
+        .edgesIgnoringSafeArea(.all)
```
sjavora, line 39: Why are we ignoring safe area edges?

Sources/Orbit/Components/Dialog.swift
```diff
@@ -6,91 +6,109 @@ import SwiftUI
...
             primaryAction: () -> Void = {},
             secondaryAction: () -> Void = {}
         )
+        case primarySecondaryAndTertiary(
+            primaryLabel: String,
+            secondaryLabel: String,
+            tertiaryLabel: String,
+            primaryAction: () -> Void = {},
+            secondaryAction: () -> Void = {},
+            tertiaryAction: () -> Void = {}
```
sjavora, line 93: Maybe a `struct` could wrap each "button"? If we end up adding another property, it will be one more parameter per button... let's say we add accessibility identifiers... that would be 9 parameters in the this init.

```swift
struct Button {
    let title: String
    let action: () async -> Void = {}
    let accessibilityIdentifier: String
}
```
(but public, of course)

This would also group related properties instead of having them interleaved like it is now. I get the feeling that these parameters are optimized for the previews here in the package... in practice you would use all of them...

Sources/Orbit/Components/Dialog.swift
PavelHolec: Because this component is meant to always cover the whole screen modally (including tabbars etc)

Sources/Orbit/Components/Dialog.swift
sjavora: `Button` might not be the best name though.

sjavora: Looking at it now, maybe we could dispense with `Buttons` and just have a variadic `init` :trollface:

PavelHolec:
> Ah shoot, let those awfully designed buttons strike again… 😒
> 
> Can we actually, really use our strongest possible force to eliminate these no-shape-looking buttons, even `ButtonLink`s in the worst case? If nothing else, … this should be a no-go.
> 
> Taking it carefully, I'd really like to push it so that we have full-width buttons in popup views so that it mimics what's designed by the platform, but I guess that'd be too much…or wouldn't it? 💡

We had this discussion today also with Hrach and we suggested some of similar points back. None were accepted. There are worse issues from my side - for example, in critical style, `all` buttons are critical, so it is even harder to distinguish the "cancel" action from others 🤷

Sources/Orbit/Components/Button.swift
```diff
@@ -283,6 +283,22 @@ extension Button {
             }
         }
     }
+
+    public struct Content: ExpressibleByStringLiteral {
```
sjavora, line 287: Do we want this conformance to be public? Will it be useful outside of previews?

sjavora: Oh wait, there is no such thing as an "internal conformance" is there 🤔

PavelHolec: Good point - added as internal conformance through extension

Sources/Orbit/Components/Dialog.swift
PavelHolec: I read it after the update - moved that UI type to Button itself (possibly adding icon etc in future) to carry button UI type into these components that have buttons themselves. (to have `Button.Content` match `Icon.Content`)

sjavora: But will these buttons support icons? 🤔

PavelHolec: They can, it depends on the wrapper - not all wrappers will use the full future set of this Content, as most wrappers define/force the "look" of those buttons and mostly just need the "data" (icon/text/id/action).

username0x0a:
> all buttons are critical

Well, that case is rather a misinterpretation of what _critical_ means. Let's consider a couple of use cases:

- starting the Booking flow, cancelling action
  - `Cancel` is critical, `Continue` is normal
- ending the Booking flow, about to submit it
  - `Confirm` is critical, other, non-quitting button is normal
- deleting user account
  - `Delete` is critical, `Cancel` button is normal

In general, _critical_ should be reserved only for destructive operations (either dropping something unfinished that can't be restored, od deleting/removing/quitting something for good).

As known, iOS splits these cases into `normal`, `cancel`, `destructive`, you can hardly ever do it wrong I'd say. 🤔

What's the use case for 2+ critical buttons?

> …some of similar points back. None were accepted.

What's _some of similar points_? I see. And I'd say, in this case specifically, it's not a question of acceptance or non-acceptance. I mean – for vital, platform-dependent decisions, this should be a question of asking US whether the design is acceptable or not. I don't really get this position of _maybe letting us make some comments and some design gods above everything will be the only source of truth_. The less I'm getting it if we stick with saying, well, we tried, and failed. I could only imagine whether _we_ haven't tried hard enough or didn't try at all because of different opinions. I'd say I can guess Hrach's position here. 😄

BTW ad the concept of primary/secondary/etc. action levels – there's no such thing. In a single alert, the options provided are **alternatives** to each other. Usually mutually exclusive, definitely not like _this action is more important than another_. Understanding it as primary/secondary is incorrect in the first place. `UIAlertController` reflects this, provides action semantics for the case. Based on this, all actions should have a very similar visual, only slightly altered based on those semantics. Not using a completely different component. Even in **Profile** screen with a listing of differently important actions, we use the same element, only slightly highlighting _Sign out_ so it looks more critical.

PavelHolec:
> > all buttons are critical
> 
> Well, that case is rather a misinterpretation of what _critical_ means. Let's consider a couple of use cases:
> 
> * starting the Booking flow, cancelling action
>   
>   * `Cancel` is critical, `Continue` is normal
> * ending the Booking flow, about to submit it
>   
>   * `Confirm` is critical, other, non-quitting button is normal
> * deleting user account
>   
>   * `Delete` is critical, `Cancel` button is normal
> 
> In general, _critical_ should be reserved for destructive operations only (either dropping something unfinished that can't be restored, od deleting/removing/quitting something for good).
> 
> As known, iOS splits these cases into `default`, `cancel`, `destructive`, you can hardly ever do it wrong I'd say. 🤔
> 
> What's the use case for 2+ critical buttons? But beware, let's not mix _critical_ with _primary_/_secondary_. ☝️
> 
> > …some of similar points back. None were accepted.
> 
> What's _some of similar points_? I see. And I'd say, in this case specifically, it's not a question of acceptance or non-acceptance. I mean – for vital, platform-dependent decisions, this should be a question of asking US whether the design is acceptable or not. I don't really get this position of _maybe letting us make some comments and some design gods above everything will be the only source of truth_. The less I'm getting it if we stick with saying, well, we tried, and failed. I could only imagine whether _we_ haven't tried hard enough or didn't try at all because of different opinions. I'd say I can guess Hrach's position here. 😄
> 
> BTW ad the concept of primary/secondary/etc. action levels – there's no such thing. In a single alert, the options provided are **alternatives** to each other. Usually mutually exclusive, definitely not like _this action is more important than another_. It's a `Yes`/`No` decision, not a list of more and less important options. Understanding it as primary/secondary is incorrect in the first place.
> 
> `UIAlertController` reflects this, provides action semantics for the case. Based on this, all actions should have a very similar visual, only slightly altered based on those semantics. Not using a completely different component.
> 
> Even if that would be a case (not a good use of alert controller then) – in **Profile** screen with a listing of differently important actions, we use the same cell elements, only slightly highlighting _Sign out_ so it looks more critical.

This repo is probably not a place to discuss this. Please put those comments to figma so designers can answer them.

sjavora (approved):

PR #91, opened 2022-04-05
Added KeyValue component

PavelHolec:
image

Generic wrapper: 

image

### DIFF
--- a/Sources/Orbit/Support/Forms/KeyValueField.swift
+++ b/Sources/Orbit/Support/Forms/KeyValueField.swift
@@ -1,0 +1,69 @@
+import SwiftUI
+
+/// A `KeyValue` container with generic content.
+///
+/// - Related components:
+///   - ``InputField``
+///   - ``Text``
+///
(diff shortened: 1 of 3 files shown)


Sources/Orbit/Support/Forms/KeyValueField.swift
```diff
@@ -0,0 +1,69 @@
+import SwiftUI
+
+/// A `KeyValue` container with generic content.
+///
+/// - Related components:
+///   - ``InputField``
+///   - ``Text``
+///
+/// - Note: [Orbit definition](https://orbit.kiwi/components/keyvalue/)
+public struct KeyValueField<Content: View>: View {
```
sjavora, line 10: Why not make this be the `KeyValue` and just have a special `init` for the case where content is a `Text`?

PavelHolec: I wanted to avoid generic content components to be at high level, so that we can put more types inside.

sjavora (approved):

PR #92, opened 2022-04-05
Update HorizontalScroll component

PavelHolec:
image

image

Pagination:
https://user-images.githubusercontent.com/35844477/161776206-0f5f3329-ee2b-42ad-bd5d-266270f1511d.mov

### DIFF
--- a/Sources/Orbit/Components/HorizontalScroll.swift
+++ b/Sources/Orbit/Components/HorizontalScroll.swift
@@ -16,6 +16,15 @@ public enum HorizontalScrollItemHeight {
     case custom(CGFloat)
 }
 
+struct HorizontalScrollWidthKey: PreferenceKey {
+
+    static var defaultValue: CGFloat { 10 }
+
+    static func reduce(value: inout CGFloat, nextValue: () -> CGFloat) {
+        value = max(value, nextValue())
+    }
+}
+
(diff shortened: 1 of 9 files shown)


Sources/Orbit/Components/HorizontalScroll.swift
```diff
@@ -16,6 +16,15 @@ public enum HorizontalScrollItemHeight {
     case custom(CGFloat)
 }
 
+struct HorizontalScrollWidthKey: PreferenceKey {
```
sjavora, line 19: This looks like an implementation details, but it's sandwiched between `public` types.

PavelHolec: Where to put it though - public goes first - and they cannot be nested in the generic type.

PavelHolec: Where to put them though. They cannot go inside the component, so I at least ordered it by public first.

sjavora: Above previews?

PavelHolec: Updated - removed the specific key and unified with others.

sjavora (approved):

PR #97, opened 2022-04-06
Added Tabs component

PavelHolec:
image

Figma:

https://www.figma.com/file/QJHiE6nSvHKm4ttskbQyrw/SegmentedControl-exploration?node-id=639%3A1945


Sources/Orbit/Support/Views/Tab.swift
```diff
@@ -0,0 +1,126 @@
...
+/// An Orbit sub-component for constructing ``Tabs`` component.
+///
+/// - Important: The Tab can only be used as a part of ``Tabs`` component, not standalone.
+public struct Tab: View {
+
+    let label: String
+    let style: TabStyle
+
+    public var body: some View {
+        // FIXME: Convert to Button with .title4 style for a background touch feedback
```
sjavora, line 46: When will this be fixed?

Sources/Orbit/Components/Tabs.swift
```diff
@@ -0,0 +1,255 @@
...
+            .accessibility(addTraits: .isButton)
+            .onTapGesture {
+                withAnimation(.easeOut(duration: 0.25)) {
+                    selectedIndex = index
+                }
+            }
+    }
+
+    @ViewBuilder func separator(index: Int, lastIndex: Int) -> some View {
+        if 0 ..< lastIndex ~= index, [selectedIndex, selectedIndex - 1].contains(index) == false {
```
sjavora, line 89: This has two `contains`, but one is spelled using `~=`... wouldn't it be easier to read if both were spelled the same?

Sources/Orbit/Components/Tabs.swift
```diff
@@ -0,0 +1,255 @@
...
+                    .stroke(Color.cloudNormal, lineWidth: BorderWidth.thin)
+            )
+    }
+
+    @ViewBuilder func tabs(for preferences: Tab.PreferenceKey.Value, in geometry: GeometryProxy) -> some View {
+        HStack(alignment: .center, spacing: 0) {
+            ForEach(preferences.indices, id: \.self) { index in
+                tab(
+                    index,
+                    lastIndex: preferences.indices.count - 1,
```
sjavora, line 63: `indices` is not necessary.

You could probably alternatively spell this `preferences.endIndex - 1` to avoid calling `count`, but it probably doesn't matter.

sjavora (changes requested):

Sources/Orbit/Components/Tabs.swift
```diff
@@ -0,0 +1,255 @@
...
+///             Tab("one")
+///             Tab("two", style: .underlinedGradient(.ink))
+///             Tab("three", style: .underlined(.blueNormal))
+///         }
+///     }
+///
+/// - Note: [Orbit definition](https://orbit.kiwi/components/structure/tabs/)
+public struct Tabs<Content: View>: View {
+
+    public let radius: CGFloat = 6
```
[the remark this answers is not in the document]
PavelHolec: Good point, it was originally a nonstandard value 9 which has been recently updated.

Sources/Orbit/Support/Views/Tab.swift
```diff
@@ -0,0 +1,126 @@
...
+    public var textColor: Color {
+        switch self {
+            case .default:                            return .inkNormal
+            case .product:                              return .inkNormal
+            case .underlined(let color):                return color
+            case .underlinedGradient(let gradient):     return gradient.color
+        }
+    }
+
+    public var startColor: Color {
```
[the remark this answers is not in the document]
PavelHolec: I would offer these public as well, so that caller can align some other elements to our various calculate sizes/spacings/colors

Sources/Orbit/Support/Views/Tab.swift
PavelHolec: I will create a follow-up, as it will not be as straightforward as the note seems.

Sources/Orbit/Components/Tabs.swift
```diff
@@ -0,0 +1,255 @@
+import SwiftUI
+
+public enum TabsDistribution {
+    /// Distributes tab widths equally (potentially make some tabs having multiline text, if enabled).
```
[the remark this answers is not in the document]
PavelHolec: Removed, as that comment is more related to the `lineLimit` parameter and affects both enum cases.

sjavora (approved):

username0x0a:
This is rather nice actually! 👍

PavelHolec:
> This is rather nice actually! 👍

Thx! It took a bit of time to convince designers to have our own iOS style, but hopefully worth it :)

PR #102, opened 2022-04-11
Remove CircularPro font from package

### LINKED ISSUE #98: Remove the font files from repo

sjavora:
Resolves #98


Package.swift
```diff
@@ -1,6 +1,9 @@
 // swift-tools-version:5.5
 import PackageDescription
 
+// Enable to use bundled Circular Pro fonts (for development only)
```
PavelHolec, line 4: "for development only" can be removed, as we should allow easy bundling for anybody who has a licence to use those fonts (= not only for development), which is also a case of our app.

Maybe it could be "(for licensed usage)" or similar instead?

Similar applies to the Usage section in readme file.

Sources/Orbit/Foundation/Typography/Font.swift
```diff
@@ -1,52 +1,70 @@
 import UIKit
 import SwiftUI
 
+var orbitFontNames: [Font.Weight: String] = [:]
+
 public extension Font {
 
+    /// Fonts used for rendering text in Orbit.
+    static var orbitTextFonts: [Font.Weight: URL?] = [
```
PavelHolec, line 9: To match `orbitFontNames`, this could be just `orbitFonts`?

Package.swift
```diff
@@ -1,6 +1,9 @@
...
+var useBundledFonts = false
```
PavelHolec, line 5: Can be `let`

PavelHolec (approved):

Sources/Orbit/Foundation/Typography/Font.swift
sjavora: I suppose, I wanted to differentiate it from the icon font, but it's not necessary.

sjavora: Updated.

Package.swift
sjavora: Updated, though not sure what you want changed in the readme... the phrasing there is taken directly from the Orbit website (see the link).

README.md
```diff
@@ -40,6 +40,16 @@ Add Orbit package to your project by adding the package dependency:
 
 ## Usage
 
+0. (Optional) Set custom fonts. [Circular Pro is only licensed for Kiwi.com domain, for non-Kiwi.com projects use an alternative](https://orbit.kiwi/foundation/typography/open-source/).
+
+```swift
+Font.orbitFonts = [
+    .regular: URL(fileURLWithPath: ...),
+    .medium: URL(fileURLWithPath: ...),
+    .bold: URL(fileURLWithPath: ...)
```
PavelHolec, line 49: We could give an exact hint (to see it is a full path), such as URL(fileURLWithPath: "file://User/Joe/Developer/orbit-swift-ui/Foundation/Typography/CircularPro-Bold.otf").

README.md
```diff
@@ -40,6 +40,16 @@ Add Orbit package to your project by adding the package dependency:
 
 ## Usage
 
+0. (Optional) Set custom fonts. [Circular Pro is only licensed for Kiwi.com domain, for non-Kiwi.com projects use an alternative](https://orbit.kiwi/foundation/typography/open-source/).
```
PavelHolec, line 43: "Set Circular Pro or custom fonts. [Circular Pro is licenced font...](...)"
...
"If you omit this steps, a default SF Pro font will be used."

PavelHolec (approved):

PR #103, opened 2022-04-11
Add Skeleton component

PavelHolec:
Figma

https://www.figma.com/file/YMfJkmUMV8xiY7fgeW6QAr/%5BOrbit%5D-Native-Components?node-id=5797%3A3217

Orbit

https://orbit.kiwi/components/progress-indicators/skeleton/

image

image

image

### DIFF
--- a/Sources/Orbit/Components/Skeleton.swift
+++ b/Sources/Orbit/Components/Skeleton.swift
@@ -1,0 +1,174 @@
+import SwiftUI
+
+/// Shows content placeholder before data is loaded.
+///
+/// - Note: [Orbit definition](https://orbit.kiwi/components/progress-indicators/skeleton/)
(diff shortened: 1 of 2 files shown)


sjavora:
Hmm.. how will this be applied in practice? Since iOS 14, SwiftUI does this using the `.redacted` modifier... can't we just use that? I guess that doesn't animate by itself...

PavelHolec:
> 

Same as with other existing components, that have some native basic version in SwiftUI. This is not hiding the actual underlaying component as `.redacted` does. This component will probably be a building block for any custom loading Skeleton shape and layout.

sjavora:
But... why? We have platform native version of this that we don't need special code for...

sjavora (approved):

PR #120, opened 2022-05-12
Basic text support for dynamic type

PavelHolec:
image

### DIFF
--- a/Sources/Orbit/Components/Heading.swift
+++ b/Sources/Orbit/Components/Heading.swift
@@ -97,14 +97,16 @@ public extension Heading {
 
         var textStyle: Font.TextStyle {
             switch self {
-                case .display:          return .largeTitle
+                case .display:
+                    return .largeTitle
                 case .displaySubtitle:
                     if #available(iOS 14.0, *) {
                         return .title2
                     } else {
                         return .headline
                     }
-                case .title1:           return .title
+                case .title1:
+                    return .title
                 case .title2:
                     if #available(iOS 14.0, *) {
                         return .title2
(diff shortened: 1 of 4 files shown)


Sources/Orbit/Components/Heading.swift
```diff
@@ -94,6 +94,34 @@ public extension Heading {
...
+        var textStyle: Font.TextStyle {
+            switch self {
+                case .display:          return .largeTitle
+                case .displaySubtitle:
+                    if #available(iOS 14.0, *) {
+                        return .title2
+                    } else {
+                        return .headline
+                    }
+                case .title1:           return .title
```
sjavora, line 107: I'd probably put these on separate lines to match the cases with availability ifs... this is harder to read IMO.

Sources/Orbit/Foundation/Typography/Font.swift
```diff
@@ -24,12 +26,12 @@ public extension Font {
...
 
-        return .custom(fontName, size: size)
+        return customFont(fontName, size: size, style: style)
     }
 
     /// Creates Orbit icon font.
-    static func orbitIcon(size: CGFloat) -> Font {
-        .custom("orbit-icons", size: size)
+    static func orbitIcon(size: CGFloat, style: Font.TextStyle = .body) -> Font {
+        return customFont(Self.orbitIconFontName, size: size, style: style)
```
sjavora, line 34: Why the `return`?

Sources/Orbit/Components/Heading.swift
PavelHolec: Updated

Sources/Orbit/Foundation/Typography/Font.swift
PavelHolec: Updated

PR #124, opened 2022-05-13
Components should have minimal height not fixed

PavelHolec:
This should make the basic Dynamic Type support more useful.

### Matching the original height

image

image

image

image

### Dynamic fonts (XS and XXL)

image

image

image

image

image

image

image

image

image

image

image


Sources/Orbit/Components/Tag.swift
```diff
@@ -9,8 +9,8 @@ import SwiftUI
 /// - Note: [Orbit definition](https://orbit.kiwi/components/tag/)
 public struct Tag: View {
 
-    public static let height = Spacing.xLarge
-    public static let minWidth = Spacing.xLarge
+    public static let verticalPadding: CGFloat = 7
```
Parabak, line 12: Do we really need all these constants to be `public`?

Sources/Orbit/Support/Layout/FrameWidthAtLeastHeightModifier.swift
```diff
@@ -0,0 +1,42 @@
...
+                content
+            }
+        }
+    }
+}
+
+extension View {
+
+    /// Ensures the width to be at least the same size as height.
+    func frameWidthAtLeastHeight() -> some View {
```
Parabak, line 22: To me, it looks like that width can't be bigger than height, so maybe we can rename to `frameWidthAsHeight`?

Sources/Orbit/Components/Tag.swift
PavelHolec: Apart from more complicated alignment guides, this is the way to make clients be able to align their custom content next to Orbit content (so that they do not have to "guess" those paddings)

Sources/Orbit/Support/Layout/FrameWidthAtLeastHeightModifier.swift
PavelHolec: That would mean it is exactly as height, thats why I put `AtLeast` there - but I am open to other suggestions, the name is horrible 😄

Sources/Orbit/Components/Button.swift
```diff
@@ -25,43 +25,52 @@ public struct Button: View {
...
-                        .animation(nil)
+                    HStack(alignment: .firstTextBaseline, spacing: .xSmall) {
+                        Icon(iconContent, size: iconSize)
+                        text
+                            .padding(.vertical, size.verticalPadding)
                     }
 
                     Spacer(minLength: 0)
 
+                    TextStrut(size.textSize)
```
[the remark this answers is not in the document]
PavelHolec: When there is no text, only an icon - or any combination of them, we need to have a single fixed height when such buttons are next to each other (44pt in case of default font size). The padding around the text itself is needed too, in case of multiline text. So we need both paddings - one to enforce the minimal size (based on text size) and the other to create padding around the content.

Sources/Orbit/Foundation/Spacing/Spacing.swift
```diff
@@ -7,46 +7,61 @@ import CoreGraphics
...
-    public static let xLarge: CGFloat = 32
+    case xLarge
     /// 44 pts.
-    public static let xxLarge: CGFloat = 44
+    case xxLarge
     /// 60 pts.
-    public static let xxxLarge: CGFloat = 60
+    case xxxLarge
+
+    public var value: CGFloat {
```
sjavora, line 30: If this is `public`, why not just make it the raw value of `Spacing`? What do we get from `.value` vs `.rawValue`?

Sources/Orbit/Support/Forms/InputStyle.swift
```diff
@@ -98,7 +100,12 @@ struct InputContent<Content: View>: View {
     }
 
     private var suffixColor: Color {
-        prefixColor
+        switch (value, state) {
+            case (_, .disabled):        return .cloudDarkerActive
+            case (.none, _):            return .inkLight
```
sjavora, line 105: This is the same as the `default` case so it could be deleted.

Sources/Orbit/Components/Button.swift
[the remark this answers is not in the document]
Parabak: Is it possible to have a multiline label on the button? Previews for some reason don't work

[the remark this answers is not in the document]
PavelHolec: It is possible, yes, we will need to support it with bigger font sizes in future.

Previews do not work by opening the package?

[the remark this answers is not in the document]
Parabak: I 💯 sure it's something local, there is no error just out of time error. Will try with clean build

Sources/Orbit/Components/Tile.swift
```diff
@@ -23,6 +36,15 @@ public struct TileButtonStyle: SwiftUI.ButtonStyle {
             case (.none, false):                        return .whiteNormal
         }
     }
+
+    func shadow(isPressed: Bool) -> TileBorderModifier.Shadow {
+        switch (isSelected, isPressed) {
+            case (false, false):    return .small
+            case (false, true):     return .default
+            case (true, false):     return .small
+            case (true, true):      return .default
```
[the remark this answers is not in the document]
PavelHolec: And thanks to that, we even fixed the switched shadow 👍  There should be no shadow when pressed.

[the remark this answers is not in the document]
PavelHolec: https://user-images.githubusercontent.com/35844477/168833480-e2c3739b-0bbc-4040-8cfb-742cedc8020f.mov

Sources/Orbit/Foundation/Spacing/Spacing.swift
PavelHolec: Now I remember - it was originally due to multiple "cases" that had the same value. Updated 👍

sjavora: I think I even saw something like this used for such cases...

```swift
enum Foo: Int {
    case someValue = 2
    case anotherValue = 3

    static var sameAsTwo: Foo { .someValue }
}
```
... but if we don't need that anymore, even better.

Sources/Orbit/Support/Forms/InputStyle.swift
```diff
@@ -91,14 +93,17 @@ struct InputContent<Content: View>: View {
...
             default:                    return .inkNormal
         }
     }
 
     private var suffixColor: Color {
-        prefixColor
+        switch (value, state) {
+            case (_, .disabled):        return .cloudDarkerActive
+            case (_, .modified):        return .blueDark
+            default:                    return .inkLight
```
sjavora, line 105: I suppose `value` can be removed from the switch now..

PavelHolec: True, updated for both properties

Sources/Orbit/Foundation/Spacing/Spacing.swift
PavelHolec: True, that could be done if we ever need it again. But for now, it can be a simple enum

sjavora (approved):

PR #132, opened 2022-05-18
Run the design token update once a week on Monday

PavelHolec:
As we link to the most up-to-date, often sometimes WIP versions of tokens, it might be better to only update once a week.

In the last week, there was an illustration change going back and forth every day.

### DIFF
--- a/.github/workflows/tokens-updater.yml
+++ b/.github/workflows/tokens-updater.yml
@@ -2,7 +2,7 @@ name: Design Tokens Updater
 
 on:
   schedule:
-    - cron: '0 6 * * 1-5' # every work-day at 6:00am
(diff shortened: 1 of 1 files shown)


.github/workflows/tokens-updater.yml
```diff
@@ -2,7 +2,7 @@ name: Design Tokens Updater
 
 on:
   schedule:
-    - cron: '0 6 * * 1-5' # every work-day at 6:00am
+    - cron: '0 0 0 ? * MON *' # every Monday
```
sjavora, line 5: According to crontab.guru, that is not valid... you want something like `0 0 * * 1`

sjavora (approved):

PR #142, opened 2022-05-25
Add custom header content to ListChoice

PavelHolec:
image

image

image

image

image

image


Sources/Orbit/Components/ListChoice.swift
```diff
@@ -163,19 +174,21 @@ public extension ListChoice {
...
         _ title: String = "",
         description: String = "",
-        icon: Icon.Content,
+        iconContent: Icon.Content,
         disclosure: ListChoiceDisclosure = .disclosure(),
         showSeparator: Bool = true,
         action: @escaping () -> Void = {},
-        @ViewBuilder content: @escaping () -> Content
+        @ViewBuilder content: @escaping () -> Content,
+        @ViewBuilder headerContent: @escaping () -> HeaderContent
```
sjavora, line 182: Should `content` precede `headerContent`? I'd expect it to be the other way around...

Sources/Orbit/Components/ListChoice.swift
```diff
@@ -326,7 +443,11 @@ struct ListChoicePreviews: PreviewProvider {
     }
 
     static var standalone: some View {
-        ListChoice(title, description: description, icon: .grid, value: value)
+        ListChoice(title, description: description, icon: .grid) {
+            customContentPlaceholder
+        } headerContent: {
```
sjavora, line 448: Ugh. I still hate this syntax.

Sources/Orbit/Components/ListChoice.swift
PavelHolec: It is based on the later SwiftUI, where API has been changed in favor of putting the "more important" closure first, so that it can be used as a first, unnamed, closure. The API change can be seen for example in `SwiftUI.Section` component.

Deprecated:
image

New:
image

Sources/Orbit/Components/ListChoice.swift
sjavora: If they at least allowed a name for the first argument as well 🤦

sjavora (approved):

PR #143, opened 2022-05-26
Support custom formatting in InputField

Parabak:
It's a draft implementation of how we might expose SwiftUI formatting API to end users.

Alternative to https://github.com/kiwicom/orbit-swiftui/pull/145

The proposal based on enum which has huge limitation.  Some API methods are uses @available attributes,  but as far as I know `enum case` doesn't support it which means
```
enum Mode {
case formatter(formatter: Formatter) // Formatter is available from iOS13
case formatStyle(style: FormatStyle) // FormatStyle is available from iOS15
```

will not compile. Do we have other options with an enum to expose different types? 

I will prepare another MR where will try to avoid such limitation with a `factoryMethod`

### DIFF
--- a/Sources/Orbit/Components/InputField.swift
+++ b/Sources/Orbit/Components/InputField.swift
@@ -13,6 +13,11 @@ import UIKit
 /// - Important: Component expands horizontally to infinity.
 public struct InputField: View {
 
+    private enum Mode {
+        case actionsHandler(onEditingChanged: (Bool) -> Void, onCommit: () -> Void, isSecure: Bool)
+        case formatter(formatter: Formatter)
+    }
+
     @Binding private var value: String
     @Binding private var messageHeight: CGFloat
     @State private var isEditing: Bool = false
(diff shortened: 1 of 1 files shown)


Sources/Orbit/Components/InputField.swift
```diff
@@ -71,42 +75,18 @@ public struct InputField: View {
...
-            textField
+        switch mode {
+            case .actionsHandler(let onEditingChanged, let onCommit, let isSecure):
+                if isSecure {
+                    secureField(onEditingChanged: onEditingChanged, onCommit: onCommit)
+                } else {
+                    textField(onEditingChanged: onEditingChanged, onCommit: onCommit)
+                }
+            case .formatter(let formatter):
+                TextField("", value: $value, formatter: formatter)
```
PavelHolec, line 86: According to docs, the formatter will be used when the field value is confirmed (probably the `onCommit` phase, by enter key, keyboard dismiss...). 

In that case, when it would not work in realtime, is there still any advantage from simply letting caller do whatever formatting it needs at the `onCommit` phase?

Sources/Orbit/Components/InputField.swift
```diff
@@ -173,6 +197,90 @@ public extension InputField {
...
+        value: Binding<String>,
+        prefix: Icon.Content = .none,
+        suffix: Icon.Content = .none,
+        placeholder: String = "",
+        state: InputState = .default,
+        textContent: UITextContentType? = nil,
+        keyboard: UIKeyboardType = .default,
+        autocapitalization: UITextAutocapitalizationType = .none,
+        isAutocompleteEnabled: Bool = false,
+        passwordStrength: PasswordStrengthIndicator.PasswordStrength = .empty,
```
PavelHolec, line 241: The formatter variant probably does not need passwordStrength as it does not support secure field

PavelHolec (approved):

PR #144, opened 2022-05-27
Unify `Icon.Content` and `Icon.Symbol` enums usage

PavelHolec:
Rechecked the usage of custom Icon content in other components. 

Updated generating scripts.

image

image

image

image

image

image

image

image


Sources/Orbit/Components/Icon.swift
```diff
@@ -43,8 +48,13 @@ public struct Icon: View {
                     .accessibility(hidden: true)
             case .countryFlag(let countryCode):
                 CountryFlag(countryCode, size: size)
-            case .sfSymbol(let systemName):
+            case .sfSymbol(let systemName, let color?):
+                Image(systemName: systemName)
+                    .foregroundColor(color)
+                    .font(.system(size: size.value * Self.sfSymbolToOrbitSizeRatio * sizeCategory.ratio))
```
sjavora, line 54: No accessibility for SF Symbols?

Sources/Orbit/Components/Icon.swift
```diff
@@ -79,11 +83,11 @@ public extension Icon {
         self.size = size
     }
     
-    /// Creates Orbit Icon component for provided SF Symbol.
+    /// Creates Orbit Icon component for provided SF Symbol with specified color.
     ///
     /// Color is configurable using `foregroundColor` modifier.
-    init(sfSymbol: String, size: Size = .normal) {
-        self.content = .sfSymbol(sfSymbol)
+    init(sfSymbol: String, size: Size = .normal, color: Color? = .inkNormal) {
```
sjavora, line 89: Does it have to be optional if it has a default non-nil value?

Sources/Orbit/Foundation/Icons/Icon.Content+Extensions.swift
```diff
@@ -0,0 +1,535 @@
+
+// Generated and updated by 'Automation/update_icons.py'
```
sjavora, line 2: Is the "and updated" bit important?

Sources/Orbit/Foundation/Icons/Icons.swift
```diff
@@ -1,277 +1,542 @@
-/// Generated and updated by 'Automation/orbit/update_icons.py'
-// swiftlint:disable:previous orphaned_doc_comment
 
+// Generated and updated by 'Automation/update_icons.py'
 public extension Icon {
 
     enum Symbol: Comparable, CaseIterable {
```
sjavora, line 5: I know it's not new, but what is the `Comparable` conformance for?

sjavora (commented):
Nice reduction in inits 👍

Sources/Orbit/Components/Icon.swift
PavelHolec: The SF Symbols already support accessibility. (can be overridden by caller if the native version is not appropriate)

Sources/Orbit/Components/Icon.swift
PavelHolec: I was thinking about making these shortcut inits with non-nil value, but the fact that they support override colour using `foregroundColor` would be more hidden and limited only to that full verbose init. 

Tried to make the API more similar to `Text` in this regard (although allowing for the override, which is not supported by `Text/Heading` component (yet), as there it may be a bit more limited to support it).

Will update the comments though.

Sources/Orbit/Foundation/Icons/Icons.swift
PavelHolec: Removed, not needed (can be later provided by caller)

Sources/Orbit/Foundation/Icons/Icon.Content+Extensions.swift
PavelHolec: 👍 Removed

Sources/Orbit/Components/Icon.swift
PavelHolec: That would limit the caller to use the verbose init whenever he would like to override the color using `.foregroundColor()`. As the `Text` component already supports this in a same way, I updated it so both components have more consistent API.

Updated the comments to be more clear and consistent with `Text`.

sjavora (approved):

PR #154, opened 2022-06-02
Add accessibility identifiers to wrapping component parts

PavelHolec:


Sources/Orbit/Support/Accessibility/AccessibilityId.swift
```diff
@@ -0,0 +1,50 @@
+import SwiftUI
+
+/// Orbit accessibility identifier for use in wrapper components.
+public enum AccessibilityId: String {
```
sjavora, line 4:
```suggestion
public enum AccessibilityID: String {
```
pls

Sources/Orbit/Support/Accessibility/AccessibilityId.swift
```diff
@@ -0,0 +1,50 @@
...
+    case keyValueValue                  = "orbit.keyvalue.value"
+    case listChoiceTitle                = "orbit.listchoice.title"
+    case listChoiceDescription          = "orbit.listchoice.description"
+    case passwordStrengthIndicator      = "orbit.passwordstrengthindicator"
+    case radioTitle                     = "orbit.radio.title"
+    case radioDescription               = "orbit.radio.description"
+    case tileTitle                      = "orbit.tile.title"
+    case tileDescription                = "orbit.tile.description"
+    case tileDisclosureButtonLink       = "orbit.tile.disclosure.buttonlink"
+    case tileDisclosureIcon             = "orbit.tile.disclosure.icon"
```
sjavora, line 41: The raw values are public if done this way. Don't we want to make them internal by moving them into a computed `var`?

sjavora (changes requested):

Sources/Orbit/Support/Accessibility/AccessibilityId.swift
PavelHolec: This was intentional, so that caller in UI test code (or PageObject creator) can directly use these to define selectors and assertions.

sjavora (approved):

PR #174, opened 2022-06-13
Refactor custom content evaluation

PavelHolec:


Sources/Orbit/Support/Forms/LazyVStackIfAvailable.swift
```diff
@@ -1,35 +1,31 @@
 import SwiftUI
 
-struct LazyVStackIfAvailable<Content: View>: View {
+public struct LazyVStackIfAvailable<Content: View>: View {
```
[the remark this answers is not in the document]
PavelHolec: It's true, but will revert it anyway and remove the init.

Sources/Orbit/Support/Layout/LazyVStackIfAvailable.swift
```diff
@@ -2,37 +2,23 @@ import SwiftUI
 
 struct LazyVStackIfAvailable<Content: View>: View {
```
username0x0a, line 3: I don't think it's even a good idea to provide this specific wrapper at all. 🤔

Firstly, the naming isn't the luckiest part as _if available_ can be understood differently, be it even _if it's not available, don't show anything at all_. _PossiblyVStack` would at least reflect the actual result correctly, altho I'm not taking the copy for this name… 😄

Secondly, the pessimistic result might do more damage than benefits – consider someone attempting to render a collection of 1000 list items, having the hope of this component doing at least something a bit useful when laziness is not available not to dump the CPU completely. I'd rather let the consumers solve this explicitly in their separate codebase as the optimisation that could be considered would be f.e. paged content or something that will definitely not calculate and “draw” 1000 elements when there's no `LazyXStack` around.

PavelHolec: Thx for the idea, it hinted me to the right direction! Renamed simply to `LazyVStack`.

username0x0a: …that still gives a false promise if the runtime platform doesn't match the minimum where the lazy counterpart is actually available. 🤔 What do we use this for exactly? I'd really prefer letting the end-user solve this issue on his own, considering different option, as we can't really say _hey, it's okay to drop some big collection on anything using this_ – that could probably only result to finding out it's not performing well and doing some `if-else`ing anyway.

sjavora (approved):

PR #177, opened 2022-06-15
Provide default ListItem init with specific sizing of dot icon.

PavelHolec:
The default `ListItem` in Orbit has a non-matching icon-text sizes which should be supported.

image

image


PavelHolec:
> Sorry about the grammar nazism, but these will become more visible as DocC is used 😉

Of course, thx for that 👍

PR #180, opened 2022-06-15
Update README.md

PavelHolec:
Could not get github markup to center image and text next to each other (html `style` attribute gets stripped) 🤷‍♂️ 

image


PavelHolec:
> So is the HTML actually needed for anything? We could just reference the image in Markdown, right?

Updated and sized the image to match the `DocC` text width to align better, but the `<br>` needs to be added, otherwise the text will be either on the same line as image, or there would be a paragraph space.

sjavora (approved):

PR #194, opened 2022-06-24
Restrict `NotificationBadge` to always be a circle

### LINKED ISSUE #159: Limit NotificationBadge shape to circle
According to Orbit, this version should be only circular.

sjavora:
It's no longer possible to have both icon and text - this matches the [orbit.kiwi definition](https://orbit.kiwi/components/information/notificationbadge/).

Screenshot 2022-06-24 at 11 31 43

Overflow behavior is also the same - the text is simply shown outside of the circle.

Screenshot 2022-06-24 at 11 32 00

Resolves #159.


PavelHolec:
👍  There is only one potential issue. We use this smaller version in Kiwi.com app, where this will probably need to be replaced by a similar custom component after this change.

Previously, the Badge had a size property in order to cover that.

Suggestion: maybe a property `allowNonCircularSize: Bool = false` maybe? 🤔

sjavora:
Sounds like misuse to me... is the app using it to display both elements at once or is it just for text?

PavelHolec (approved):

PavelHolec:
> Sounds like misuse to me... is the app using it to display both elements at once or is it just for text?

We should limit to use either - this means, the init should take either `String` or `Icon.Content`.

The Orbit specifies the String content as limited to 2 characters, so what we can additionally do to make it more flexible is:
- when text.length <=2, enforce circle
- otherwise, let the shape resize with the text

This is currently what Orbit says:
> Use only an icon or a number no more than 2 digits.

Although limiting the text to `Int` only would not be flexible in the long run.

sjavora:
This issue is literally about only allowing circular badges... it seems bonkers to me to go "ok, let's do that, but also let's not in case it's not convenient"... why bother at all then?

PR #201, opened 2022-06-24
CountryFlag: fallback to unknown flag

### LINKED ISSUE #86: CountryFlag: fallback to unknown flag
As in

https://github.com/kiwicom/orbit-compose/pull/127/commits/3485ccf486ff8a038ff02bba60159e8c8fd6d14d

sjavora:
Resolves #86.


PavelHolec (approved):

username0x0a:
I remember drawing that greyscale-striped image using the Circular font we use in app, I wonder where that raster ended 🤔 but I could reproduce if we'd like to align it. 😄

hrach:
@username0x0a shut up and take my money and create the PR.

Sources/Orbit/Components/CountryFlag.swift
```diff
@@ -28,6 +28,10 @@ public struct CountryFlag: View {
         }
     }
 
+    var imageName: String {
+        UIImage(named: countryCode.lowercased(), in: .current, with: nil) != nil ? countryCode.lowercased() : "unknown"
```
sjavora, line 32: Even though I'm sure this is cached, it's still unfortunate that we need to try to create an image just to see if it is bundled... #63 might help with this 🤔

PavelHolec (approved):

PR #203, opened 2022-06-28
Remove clear button from InputField

PavelHolec:


sjavora (approved):
Don't we want a clear button anymore?

PavelHolec:
> Don't we want a clear button anymore?

Orbit suggests not to have it. Android does not have it too. But, if client wants, a suffix still can be used for this.

PR #215, opened 2022-06-29
Support enum type for `CountryFlag`

### LINKED ISSUE #63: Support enum type for CountryFlag

sjavora:
Named it `CountryFlag.CountryCode`. I made the enum with `String` rawValue to get the free failable init - otherwise that would have to be generated as well.

Resolves #63.


Automation/update_flags.py
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env python3
```
PavelHolec, line 1: The filename could be `update_countryflags` or `update_country_flags` to match the naming (`flags` alone sound like technical/feature flags)

Sources/Orbit/Components/CountryFlag.swift
```diff
@@ -49,11 +43,19 @@ public struct CountryFlag: View {
...
     /// Creates Orbit CountryFlag component.
-    init(_ countryCode: String, size: Icon.Size = .normal, border: Border = .default()) {
+    init(_ countryCode: CountryCode, size: Icon.Size = .normal, border: Border = .default()) {
         self.countryCode = countryCode
         self.size = size
         self.border = border
     }
+
+    /// Creates Orbit CountryFlag component with a string country code.
+    /// If a corresponding image is not found, the flag for unkonwn codes is used.
```
PavelHolec, line 53: `unkonwn` typo. The second sentence might be separated by a newline.

Sources/Orbit/Components/CountryFlag.swift
```diff
@@ -86,7 +88,8 @@ struct CountryFlagPreviews: PreviewProvider {
     static var standalone: some View {
         VStack {
             CountryFlag("cz")
-            CountryFlag("")     // EmptyView
+            CountryFlag("") // unknown flag
+            CountryFlag("some invalid identifier") // unknown flag
```
PavelHolec, line 92: This would now render the unknown flag, which should go into a separate property (`var unknown`). The original behaviour was to draw nothing (EmptyView), which was the intent of the VStack. Now, the VStack can be removed and the standalone preview should only show the single `CountryFlag("cz")`.

Automation/update_flags.py
sjavora: 👍

Sources/Orbit/Components/CountryFlag.swift
sjavora: 👍

Sources/Orbit/Components/CountryFlag.swift
sjavora: 👍

PavelHolec (approved):

Sources/Orbit/Foundation/CountryFlags/CountryFlags.swift
```diff
@@ -0,0 +1,264 @@
...
+        case to
+        case tr
+        case tt
+        case tv
+        case tw
+        case tz
+        case ua
+        case ug
+        case um
+        case unknown
```
username0x0a, line 244: I'd probably keep this `case unknown` as the first member in the enumeration, as a good practise. 🤔

PR #218, opened 2022-06-30
Release updates

PavelHolec:
Updated the lvl1 elevation further.

Figma with swiftui preview overlays in the middle of rectangles

image


Sources/Orbit/Foundation/CountryFlags/CountryFlags.swift
```diff
@@ -260,5 +260,9 @@ public extension CountryFlag {
         case za
         case zm
         case zw
+
+        public init(_ string: String) {
```
sjavora, line 264: This file is autogenerated, so the change will be lost the next time flags are changed - I suggest moving this to a private extension in `CountryFlag.swift`.

Sources/Orbit/Support/Components/TileBorder.swift
```diff
@@ -97,17 +105,18 @@ public struct TileBorderModifier: ViewModifier {
...
+        }
     }
 }
 
 public extension View {
 
-    /// Decorates content with a ``Tile`` appearance using specified style.
-    ///
-    /// - Parameter style: The style to apply. If style is nil, the view doesn’t get decorated.
+    /// Decorates content with a ``Tile`` or ``Card`` appearance using specified style.
```
sjavora, line 117: If it is truly only meant for `Tile` and `Card`, shouldn't there be extensions for only those types? 🤔

sjavora (changes requested):

sjavora:
Also please mention this somewhere in the documentation so it doesn't show up on the main page...

Screenshot 2022-07-13 at 13 10 13

Sources/Orbit/Support/Components/TileBorder.swift
PavelHolec: It is more like decorates with Tile/Card-like appearance. Anyway, this modifier will be replaced in future with more composable modifiers (separate selectable border).

Sources/Orbit/Foundation/CountryFlags/CountryFlags.swift
PavelHolec: Thx, updated 👍

Sources/Orbit/Support/Components/TileBorder.swift
PavelHolec: Updated

PR #226, opened 2022-07-14
Allow multiline content in Badge

PavelHolec:
Even when some components are meant to be single line, we should not force it anymore, as it breaks the layout for dynamic types or some other edge-cases (verbose languages). Also removed the width constraint which was only needed for the dynamic size anyway.

The only last component that still had the horizontal `fixedSize` is the `Badge` component.

image

image

image


Sources/Orbit/Components/Badge.swift
```diff
@@ -36,7 +36,6 @@ public struct Badge: View {
                     .padding(.vertical, Self.verticalPadding)
             }
             .padding(.horizontal, .xSmall)
-            .frameWidthAtLeastHeight()
```
sjavora, line 39: Haven't looked into how this is implemented, but do we really want to get rid of this one?

PavelHolec: It was mostly needed for the NotificationBadge to force circular shape for narrow numbers. This modifier has issue with multiline text, so it needed to be removed here (and probably removed completely later if the badges would merge again)

sjavora (approved):

PR #229, opened 2022-07-18
Reduce compression resistance for SecureTextFied to fit wrapper

Parabak:
Fix issue with SecureTextField's growing frame.


sjavora:
Aaah, Auto Layout strikes from beyond the grave 😄

PavelHolec (approved):

PR #242, opened 2022-07-23
Update darkmode based on new experimental palette

PavelHolec:
Using few experimental dark mode colors from figma

https://www.figma.com/file/bpXcYuAuuhYAdrJoT2U7yz/Dark-mode?node-id=0%3A1

image

image

image

image


username0x0a (approved):
🤩

username0x0a:
There's a couple of padding moves, yet some snapshots seem to lose the padding completely – is that on purpose, like do we only retain it on actual Previews for human-readable cases? 🤔

PavelHolec:
> There's a couple of padding moves, yet some snapshots seem to lose the padding completely – is that on purpose, like do we only retain it on actual Previews for human-readable cases? 🤔

I will recheck the previews. In general, the Storybook-related paddings have been moved out to unified layout. Will add .medium paddings back to specific previews where they are now missing.

hrach:
So in Compose, we have those colored secondary buttons on Alert rather lighter than darker then the alert's background. Similarly the button is darker than the bg in light mode. Is this intentional?

PavelHolec:
> So in Compose, we have those colored secondary buttons on Alert rather lighter than darker then the alert's background. Similarly the button is darker than the bg in light mode. Is this intentional?

This task is using only the figma colors as direct replacement of existing palette colors. No additional surface/opacity etc. logic is added yet - waiting for more progress from Orbit.

PavelHolec:
> So in Compose, we have those colored secondary buttons on Alert rather lighter than darker then the alert's background. Similarly the button is darker than the bg in light mode. Is this intentional?

Will recheck - I probably missed some of WIP colors from figma.

hrach:
Those colors weren't present at all in Figma, if you don't have a different file. I made them 10% lighter.

PavelHolec:
> 

Did with similar offsets (also for active+hover), updated.

PavelHolec:
> Those colors weren't present at all in Figma, if you don't have a different file. I made them 10% lighter.

Did it similar with similar offsets (also for active+hover), updated.

PR #255, opened 2022-07-28
Add compact variant to `InputField`

PavelHolec:


Sources/Orbit/Components/InputField.swift
```diff
@@ -1,6 +1,14 @@
 import SwiftUI
 import UIKit
 
+/// Style variant for Orbit InputField component.
+public enum InputFieldStyle {
+
+    case `default`
```
sjavora, line 7: This should get some documentation as well.

Sources/Orbit/Support/Forms/InputState.swift
```diff
@@ -0,0 +1,38 @@
+import SwiftUI
+
+public enum InputState {
```
sjavora, line 3: No documentation here.

sjavora (changes requested):
Please check the generated docs if the root page is clean.

Sources/Orbit/Support/Forms/InputState.swift
PavelHolec: Added comment only to the enum, because it will probably be completely removed in future story (and replaced with native swiftui `disabled` and `focusable` modifiers).

sjavora:
I really wish I could find a way to automate this check...

Screenshot 2022-07-29 at 17 07 56

sjavora (changes requested):
Unsorted docs are visible again in root DocC.

PavelHolec:
@sjavora docs should be updated now.

sjavora (approved):

PR #272, opened 2022-08-16
Add ScreenLayoutPaddingKey

PavelHolec:
This allows custom components to have Card-like edge-to-edge layout.

image

image


Sources/Orbit/Support/Layout/IgnoreScreenLayoutHorizontalPaddingModifier.swift
```diff
@@ -0,0 +1,66 @@
...
+
+        return -screenLayoutPadding.horizontal(horizontalSizeClass: horizontalSizeClass)
+    }
+}
+
+public extension View {
+
+    /// Reverts any horizontal padding provided by `screenLayout` context.
+    ///
+    /// A typical usage is to mimic the edge-to-edge appearance of a `Card` component.
```
Parabak, line 26: Shall we mention in the description, that it's valid only for `compact` mode?

Sources/Orbit/Support/Layout/ScreenLayoutModifier.swift
```diff
@@ -1,72 +1,72 @@
...
+        switch self {
+            case .default:                      return `default`(horizontalSizeClass: horizontalSizeClass)
             case .compact:                      return .medium
             case .custom(_, _, let bottom):     return bottom
         }
     }
 
-    var defaultPadding: CGFloat {
+    public func `default`(horizontalSizeClass: UserInterfaceSizeClass?) -> CGFloat {
         horizontalSizeClass == .regular
```
Parabak, line 37: Suggest to use switch-case.

Sources/Orbit/Support/Layout/IgnoreScreenLayoutHorizontalPaddingModifier.swift
PavelHolec: Do you mean `compact` width (interface size) or `compact` padding (one of padding cases)?

Sources/Orbit/Support/Layout/ScreenLayoutModifier.swift
PavelHolec: It would be unnecessary larger though, with optional size class.

Sources/Orbit/Support/Layout/IgnoreScreenLayoutHorizontalPaddingModifier.swift
Parabak: I meant compact width.
We do check `horizontalSizeClass != .regular` and if it's `horizontalSizeClass == .regular` we appply zero padding.

I suggest reflecting that in  docs

PavelHolec: I am probably missing something. This modifier is independent on it, it specifies that it "reverts **any** horizontal padding", meaning "if it was applied, revert it". 
The call site cannot optionally use this modifier, it needs to use it all the time anyway.

PavelHolec: Thx, I finally get it! 👍  
I wanted to even rename it to reflect that, but found no good name. Updated the doc comments.

Parabak (approved):

PR #273, opened 2022-08-16
Remove fixed size of EmptyState button

### LINKED ISSUE #269: `EmptyState` button has fixed vertical size
The `EmptyState` now uses `fixedSize` `Button` to keep the width as needed. 

This works for small content, but makes the button grow over provided width. Applying `.fixedSize` on stack inside the `Button` does not work, as it makes the whole container fixedSize and does not respect the `Text` inside being "not" fixedSize (in horizontal axis).

#### Solution
The horizontal (and vertical) expanding property of all components will need to be configurable (disabled) by a modifier.

#### Suggestions
- modifier `.idealSize(horizontal:vertical)` (similar to `.fixedSize()`) that propagates information to subviews that they should not be expanding beyond the ideal size in the provided axis.

PavelHolec:
Closes #269 

Applied the modifier to other expanding components.

image

image

image

image


Parabak (approved):

PavelHolec:
@sjavora @Parabak please recheck - the only solution I found to address the issue is to add a new modifier (see issue description)

Sources/Orbit/Components/Button.swift
```diff
@@ -289,21 +293,26 @@ struct ButtonPreviews: PreviewProvider {
...
+            }
+            .idealSize(horizontal: true)
+
+//            standaloneCombinations
+//            sizing
+//
+//            storybook
+//            storybookStatus
+//            storybookGradient
+//            storybookMix
```
sjavora, line 308: Commented code.

Sources/Orbit/Components/Button.swift
```diff
@@ -432,32 +441,32 @@ struct ButtonPreviews: PreviewProvider {
...
+//        .padding(.medium)
+//        .previewLayout(.sizeThatFits)
+//    }
+//
+//    @ViewBuilder static var content: some View {
+//        ButtonPreviews.standaloneCombinations
+//        ButtonPreviews.sizing
+//        ButtonPreviews.buttons(.primary)
+//    }
+//}
```
sjavora, line 472: More commented code.

sjavora (changes requested):

Sources/Orbit/Components/Button.swift
PavelHolec: Thx, updated both

sjavora (approved):

PR #276, opened 2022-08-22
Add Heading and Icon concatenation support

bul-nick-al:
Add the ability to concatenate `Icon` and `Heading` of various styles in any order. Inspired by [How to insert images into text](https://www.hackingwithswift.com/quick-start/swiftui/how-to-insert-images-into-text). 

This addition will allow us to create layouts with text akin to this
Screenshot 2022-08-22 at 23 47 17


Sources/Orbit/Support/TextConcatenation/SwiftUITextRepresentable+Concatenation.swift
```diff
@@ -0,0 +1,41 @@
+import SwiftUI
+
+/// Conctatenates two terms that can be represented as `SwiftUI.Text`
+/// - Parameters:
+///   - left: An entity, usually a view, which content can be represented as `SwiftUI.Text`
+///   - right: An entity, usually a view, which content can be represented as `SwiftUI.Text`
+/// - Returns: A view that is the result of concatenation of text representation of the parameters.
+///   if both paramters do not have a text representation, the returning view will produce `EmptyView`, preserving the standard Orbit behaviour
```
sjavora, line 8: Let's stick with US English.

Sources/Orbit/Support/TextConcatenation/TextRepresentable.swift
```diff
@@ -0,0 +1,11 @@
+import SwiftUI
+
+public protocol SwiftUITextRepresentable {
+    func asText(configuration: ContentSizeCategory) -> SwiftUI.Text?
```
sjavora, line 4: Hmm I don't recall many "asSomething"s in Swift... I was originally going to write that this should be expressed as an initializer, but that would be an initializer on `SwfitUI.Text` 🤔 

Something just doesn't feel "Swifty" about this, but I can't think of a better way to do it at the moment 😄

Sources/Orbit/Support/TextConcatenation/DynamicTextWrapper.swift
```diff
@@ -0,0 +1,36 @@
...
+
+    var wrapped: DynamicTextWrapper {
+        self
+    }
+}
+
+func +(
+    left: DynamicTextWrapper,
+    right: DynamicTextWrapper
+) -> DynamicTextWrapper {
```
sjavora, line 28: The declaration comfortably fits on a single line.

Sources/Orbit/Components/Icon.swift
```diff
@@ -55,6 +55,64 @@ public struct Icon: View {
...
+            case .image, .countryFlag, .sfSymbol:
+                assertionFailure(".image, .countryFlag, .sfSymbol as Text are only available in iOS 14.0 or newer")
+                return nil
+        }
+    }
+}
+
+extension Icon: SwiftUITextRepresentable {
+    public func asText(configuration: ContentSizeCategory) -> SwiftUI.Text? {
+        if content.isEmpty == false {
```
sjavora, line 106: Since this is not a result builder, maybe let's use an early exit?

```swift
if content.isEmpty { return nil }
```

sjavora (changes requested):

sjavora:
Thinking about it more, do we even need the wrapper?

bul-nick-al:
> Thinking about it more, do we even need the wrapper? We could make a generic `+` instead...

We need a wrapper, that is a view because that seems to be the only way to pass `sizeCategory` to each of the concatenated texts. The correct way would be using `ViewModifier` for such purposes but it does not allow to preserve the view type, even if the applied modifier does not change that type. In this case I need to `SwiftUI.Text` to stay `SwiftUI.Text` and `.modifier` does not let me do that, there is a[ stack overflow question about that](https://stackoverflow.com/a/56906557)

Sources/Orbit/Support/TextConcatenation/SwiftUITextRepresentable+Concatenation.swift
bul-nick-al: Corrected

Sources/Orbit/Support/TextConcatenation/TextRepresentable.swift
bul-nick-al: renamed to `swiftUITextContent`. Can't come up with a better name

Sources/Orbit/Support/TextConcatenation/DynamicTextWrapper.swift
bul-nick-al: done

Sources/Orbit/Components/Icon.swift
bul-nick-al: replaced with `guard`

sjavora (approved):

username0x0a (approved):

sjavora:
Hmm that's quite a lot of failures 🤔 

No idea how to trigger a manual job on a fork though...

Sources/Orbit/Support/TextConcatenation/DynamicTextWrapper.swift
```diff
@@ -0,0 +1,52 @@
+import SwiftUI
+
+struct DynamicTextWrapper: View {
```
PavelHolec, line 3: Suggestion: `ConcatenatedTextWrapper` ? `Dynamic` is a vague term.

Sources/Orbit/Support/TextConcatenation/SwiftUITextRepresentable+Concatenation.swift
```diff
@@ -0,0 +1,42 @@
+import SwiftUI
+
+/// Conctatenates two terms that can be represented as `SwiftUI.Text`
+/// - Parameters:
+///   - left: An entity, usually a view, which content can be represented as `SwiftUI.Text`
```
PavelHolec, line 5: Suggestion:

- left: Content representable as `SwiftUI.Text`

Sources/Orbit/Support/TextConcatenation/SwiftUITextRepresentable+Concatenation.swift
```diff
@@ -0,0 +1,42 @@
...
+                    .padding(.vertical, .small)
+                Heading("Title 1 ", style: .title1)  + Icon(.flightReturn, size: .heading(.title1))
+                Heading("Title 2 ", style: .title2) + Icon(.flightReturn, size: .heading(.title2))
+                Heading("Title 3 ", style: .title3) + Icon(.flightReturn, size: .heading(.title3))
+                Heading("Title 4 ", style: .title4) + Icon(.flightReturn, size: .heading(.title4))
+                Heading("Title 5 ", style: .title5) + Icon(.flightReturn, size: .heading(.title5))
+                Heading("Title 6 ", style: .title6) + Icon(.flightReturn, size: .heading(.title6))
+            }
+        }
+//        .environment(\.sizeCategory, .accessibilityExtraExtraExtraLarge)
```
PavelHolec, line 37: A leftover probably

Sources/Orbit/Support/TextConcatenation/TextRepresentable.swift
```diff
@@ -0,0 +1,5 @@
+import SwiftUI
+
+public protocol SwiftUITextRepresentable {
```
PavelHolec, line 3: The filename does not match

Sources/Orbit/Support/TextConcatenation/DynamicTextWrapper.swift
bul-nick-al: Renamed to `ConcatenatedText`

Sources/Orbit/Support/TextConcatenation/SwiftUITextRepresentable+Concatenation.swift
bul-nick-al: Changed

Sources/Orbit/Support/TextConcatenation/SwiftUITextRepresentable+Concatenation.swift
bul-nick-al: Yup, removed. Plus added dynamic type previews and snapshot tests

Sources/Orbit/Support/TextConcatenation/TextRepresentable.swift
bul-nick-al: Renamed back to `TextRepresentable`

PavelHolec (approved):

PavelHolec:
> Hmm that's quite a lot of failures 🤔
> 
> No idea how to trigger a manual job on a fork though...

@sjavora @bul-nick-al it should be available for running from the **fork** repo

image

sjavora:
@PavelHolec nope.

Screenshot 2022-08-25 at 9 15 54

PavelHolec:
> @PavelHolec nope. At least not for me.

It has to be the

PavelHolec:
> @PavelHolec nope. At least not for me.

Thats because the repo owner has to first accept the workflows. And even then, it could be that only the author can see them (could also depend on the forked repo settings)

PR #277, opened 2022-08-24
Update Elevation modifier

PavelHolec:
Refactored to avoid rendering the original content with `drawingGroup` which caused errors with some content (TextViews)


Sources/Orbit/Foundation/Elevations/Elevation.swift
```diff
@@ -11,18 +11,31 @@ public enum Elevation {
     case custom(opacity: Double, radius: CGFloat, x: CGFloat = 0, y: CGFloat = 0, padding: CGFloat? = nil)
 }
 
-public extension View {
-
-    /// Elevates the view to make it more prominent.
-    func elevation(_ level: Elevation?) -> some View {
-        elevation(level, prerender: true)
-    }
+public enum ElevationShape {
```
sjavora, line 14: Missing docs on the type itself.

PavelHolec: True, updated

sjavora (approved):

PavelHolec:
@sjavora updated code and doc comments further

PR #278, opened 2022-08-25
Add `Heading` and `Icon` concatenation support

bul-nick-al:
Add the ability to concatenate `Icon` and `Heading` of various styles in any order. Inspired by [How to insert images into text](https://www.hackingwithswift.com/quick-start/swiftui/how-to-insert-images-into-text). 

This addition will allow us to create layouts with text akin to this
Screenshot 2022-08-22 at 23 47 17

Clone of [!276](https://github.com/kiwicom/orbit-swiftui/pull/276)


sjavora:
@PavelHolec what do you think about the snapshots?

PavelHolec:
> @PavelHolec what do you think about the snapshots?

All sizes seem to be off.. 🤷  I hoped this task only touches `Heading` and nothing else.

sjavora (approved):

Tests/SnapshotTests/TextConcatenationTests.swift
```diff
@@ -0,0 +1,9 @@
+import XCTest
+@testable import Orbit
+
+class TextConcatenationTests: SnapshotTestCase {
+
+    func testTextConcatenation() {
+        assert(TextConcatenationPreviews.snpashot)
```
PavelHolec, line 7: Typo `snpashot`

bul-nick-al: Oops, renamed

Sources/Orbit/Support/TextConcatenation/TextRepresentable+Concatenation.swift
```diff
@@ -0,0 +1,70 @@
...
+
+    static var standalone: some View {
+        (Heading("Hanoi ", style: .title1)
+             + Icon(.flightReturn, size: .xLarge)
+             + Heading(" San Pedro de Alcantara", style: .title1))
+        .lineLimit(2)
+    }
+    static var snapshot: some View {
+        VStack(alignment: .leading, spacing: .xSmall) {
+            Heading("Display Title", style: .display) + Icon(.flightReturn, size: .heading(.display))
```
PavelHolec, line 34: The current snapshots now use the icon size as if it would be used in a `Label` or `Tile` - but that uses specifically larger size for that case.

Suggestion: using the same size as the heading (for now using the `custom` enum, but we should add a mapping property now or in future), so that it better matches when the Icon is **inside** the text:

```
Heading(.., style) + Icon(.flightReturn, size: .custom(style.size)) + Heading(.., style)
```

Result:

image

image

bul-nick-al: Updated the snapshots

PavelHolec (approved):

PR #281, opened 2022-08-26
Make `PreviewWrapper` internal

sjavora:
In the Kiwi app, we currently define a "`PreviewWrapper`" that has a slightly different implementation (by necessity) from the one in here. The thing is that in my opinion we never actually want to use the PreviewWrapper from Orbit, since we pack extra logic into the one in the app - using this one is always a mistake.

So we can either
1. always add a module name to disambiguate - inelegant, error-prone
2. come up with a different name for the one in the app
3. make this one internal

I think the last option makes the most sense - when working on Orbit, everything works the same as before and when importing Orbit elsewhere, this wrapper will not get in the way - clients can still copy it if they want, and presumably add their own logic to it.

At first I also thought about adding an underscore before the Orbit wrapper so it could remain public while not showing up in autocomplete or documentation, but then I realized we don't actually use it in the app at all, so I think there is no point.

If this is merged, all packages in the Kiwi app will use the other version - we can even get rid of the typealias there.

### DIFF
--- a/Sources/Orbit/Orbit.docc/Uncategorized.md
+++ b/Sources/Orbit/Orbit.docc/Uncategorized.md
@@ -21,7 +21,6 @@ Supporting types and components under development.
 
 ### Xcode previews support
 
-- ``PreviewWrapper``
 - ``StateWrapper``
 
 ### SwiftUI environment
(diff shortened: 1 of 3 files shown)


PavelHolec:
Agree. Its a pity that there is no easier way to apply fonts without using a wrapper with initializer. Or at least, I have not found any.

PavelHolec (approved):

PR #287, opened 2022-08-29
Allow customizing colors

### LINKED ISSUE #268: Customizing Colors for system
Hello, I'm just wondering if there are any plans to add custom color themes. I'd like to use my "product" color that's different than the one baked in. I'm working with a local copy of the repo and changing a couple of the hues, but I'm wondering if there's a better way to go about this? Maybe I could contribute to this if it's not on the docket. I did see a "Prepare design tokens for native platforms" item on the roadmap on your website.

PavelHolec:
Closes #268 

A simpliest way would be to make colors `var`s. Not the most elegant solution, but it might be a workaround until orbit designTokens are supported.

image


sjavora:
Do we know how we'll support the tokens?

PavelHolec:
> Do we know how we'll support the tokens?

They are still WIP on Orbit side, so it is hard to tell

https://orbit.kiwi/foundation/design-tokens/

sjavora (approved):

PR #294, opened 2022-08-30
Add basic `TextRepresentable` support to `Text`

PavelHolec:
Basic support with known limitations:

- cannot control `countryFlag` size
- concatenated Icon sizing is incorrect for accessibility sizes
- concatenated Text does not support `TextLink`s (only formatting) 
- negative baseline offset on sfFonts seems to cause strange resize effect

image

image


Sources/Orbit/Components/Heading.swift
```diff
@@ -22,7 +22,7 @@ public struct Heading: View {
         }
     }
 
-    func textContent(sizeCategory: ContentSizeCategory) -> SwiftUI.Text {
+    @ViewBuilder func textContent(sizeCategory: ContentSizeCategory) -> SwiftUI.Text {
```
sjavora, line 25: What's the point of `@ViewBuilder` if we don't return `some View`?

Sources/Orbit/Components/Heading.swift
```diff
@@ -180,6 +172,16 @@ public extension Heading {
     }
 }
 
+// MARK: - TextRepresentable
+extension Heading: TextRepresentable {
+
+    public func swiftUITextContent(configuration: ContentSizeCategory) -> SwiftUI.Text? {
+        guard text.isEmpty == false else { return nil }
```
sjavora, line 179:
```swift
if text.isEmpty { return nil }
```

Sources/Orbit/Components/Text.swift
```diff
@@ -254,6 +265,16 @@ extension TextAlignment {
     }
 }
 
+// MARK: - TextRepresentable
+extension Text: TextRepresentable {
+
+    public func swiftUITextContent(configuration: ContentSizeCategory) -> SwiftUI.Text? {
+        guard content.isEmpty == false else { return nil }
```
sjavora, line 272: Same as the other file.

sjavora (changes requested):

sjavora:
@PavelHolec we should have an issue tracking the limitations outlined in the description.

Sources/Orbit/Components/Text.swift
PavelHolec: Copied from previous implementation, updated.

Sources/Orbit/Components/Heading.swift
PavelHolec: Updated

PavelHolec:
> @PavelHolec we should have an issue tracking the limitations outlined in the description.

Added as #295

sjavora (approved):

Sources/Orbit/Components/Heading.swift
bul-nick-al: Just out of curiosity, why is it better? When `guard` was introduced, such cases were given as an example where it should be used 🤔

sjavora: @bul-nick-al it's shorter _and_ has no negation. If it weren't for the rule of not using `!`, we'd have
```swift
guard !text.isEmpty else { return nil }
```
...which my brain just refuses to process. Still, no negation is better than readable negation. It's not a bias against `guard`, either. If we were checking for emptiness instead, I'd happily use `guard text.isEmpty`.

PR #327, opened 2022-10-10
Update CI to use Xcode 14

### LINKED ISSUE #320: Move snapshots to Xcode 14/iOS 16
- update jobs to generate and test snapshots to use Xcode 14
- regenerate snapshots

sjavora:
~Same changes to simulators as in Kiwi - SE goes from 1st gen to 3rd gen.~

Simplified to iPhone and iPad.

Resolves #320.


username0x0a (approved):

PavelHolec:
I am thinking whether it would be better to not include the iOS version in the folder name and have simply folders like `iPad`, `iPhone SE`, `iPhone 13`. Because with each iOS change the diff will contain only all deleted and all new files.

sjavora:
Ok, so updated so that only two folders of snapshots exist:
- iPhone
- iPad

Triggered regeneration job...

PavelHolec (approved):

PR #345, opened 2022-10-28
OrbitStorybook target

### LINKED ISSUE #323: Extract storybook and other demo related code to separate target
Extract code that is mostly needed only for Orbit Storybook app to separate target so that the main target is faster to build.

sjavora:
Previews prefixed with "storybook" that were used in the Storybook view were copied over to the new OrbitStorybook target as new types (akin to StorybookTypography).

This means that Orbit's previews get to stay internal, but can now go out of sync with what is in the storybook. According to @PavelHolec, this isn't a problem since each serves a different purpose/audience.

Closes #323.


PavelHolec:
For the previews - if it is only few of them, we could just "duplicate" the layout they both share. In future, they might differ on purpose, so they do not need to be necessarily linked.

Package.swift
```diff
@@ -11,6 +11,7 @@ let package = Package(
         .library(name: "Orbit", targets: ["Orbit"]),
         .library(name: "OrbitDynamic", type: .dynamic, targets: ["Orbit"]),
         .library(name: "OrbitStatic", type: .static, targets: ["Orbit"]),
+        .library(name: "OrbitStorybook", targets: ["OrbitStorybook"]),
```
PavelHolec, line 14: In a test project that uses both Orbit and OrbitStorybook now, there are probably some naming collisions. I will try to dig deeper, but maybe you will have an idea why it might happen:

image

sjavora:
Each component has at least one "storybook" preview, but usually there are more. As I wrote above, some are even used for snapshots. I'm ok with moving those to the other target, but not sure I'd want to duplicate them. Wouldn't that be confusing? At least we should then rename those original ones so they don't mention "Storybook" anymore?

sjavora:
Possibly related, but can we get rid of the explicit static/dynamic targets?

Package.swift
sjavora: I tried that with test app and this didn't happen. 

Screenshot 2022-10-29 at 10 45 04

Screenshot 2022-10-29 at 10 44 45

sjavora:
> For the previews - if it is only few of them

It's everyting in [this commit](https://github.com/kiwicom/orbit-swiftui/pull/345/commits/062e3c167ff1fcfc30874ab5738f39f3be6e0c70). Please let me know how you'd like to proceed.

Package.swift
PavelHolec: Probably something wrong in my project

PavelHolec:
> Possibly related, but can we get rid of the explicit static/dynamic targets?

I thought about that, but potential Xcode11+ iOS13 clients might need it to solve similar issues we had before

PavelHolec:
> Each component has at least one "storybook" preview, but usually there are more. As I wrote above, some are even used for snapshots. I'm ok with moving those to the other target, but not sure I'd want to duplicate them. Wouldn't that be confusing? At least we should then rename those original ones so they don't mention "Storybook" anymore?

Yep, we could remove all traces of "storybook" there. It is possible our needs for snapshot content and storybook content might differ in future, so they do not need to be related.

sjavora:
> > Possibly related, but can we get rid of the explicit static/dynamic targets?
> 
> I thought about that, but potential Xcode11+ iOS13 clients might need it to solve similar issues we had before

I find this "support" for Xcode11+ / iOS 13 to be... weird. We're not actually checking that everything works with those versions (manually or in CI) and I have a feeling the storybook target will not work for precisely the reason that we have those static/dynamic targets.

sjavora:
> > Each component has at least one "storybook" preview, but usually there are more. As I wrote above, some are even used for snapshots. I'm ok with moving those to the other target, but not sure I'd want to duplicate them. Wouldn't that be confusing? At least we should then rename those original ones so they don't mention "Storybook" anymore?
> 
> Yep, we could remove all traces of "storybook" there.

What does that mean? Renaming them, but keeping them in place, or only moving them to the storybook target?

sjavora:
@PavelHolec updated, see new description. One thing I'm still not sure about - do you want to remove the "storybook"-prefixed previews from Orbit, or should they stay for now?

PavelHolec:
> @PavelHolec updated, see new description. One thing I'm still not sure about - do you want to remove the "storybook"-prefixed previews from Orbit, or should they stay for now?

Not necessarily now, but in future, I assume they will be removed/renamed, as the goals of each would be a bit different.

Internal Orbit previews/snapshots:
- aimed more for development
- containing all edge-cases we support (different, multi-line and rich text content)

Storybook content:
- "happy path" state of components, ideally constructed 1:1 to (Figma) designs and matching its (often simple) texts

PavelHolec:
> > > Possibly related, but can we get rid of the explicit static/dynamic targets?
> > 
> > 
> > I thought about that, but potential Xcode11+ iOS13 clients might need it to solve similar issues we had before
> 
> I find this "support" for Xcode11+ / iOS 13 to be... weird. We're not actually checking that everything works with those versions (manually or in CI) and I have a feeling the storybook target will not work for precisely the reason that we have those static/dynamic targets.

Let's remove it for now. We can always easily re-add it, if anybody would truly need it

sjavora:
> > > > Possibly related, but can we get rid of the explicit static/dynamic targets?
> > > 
> > > 
> > > I thought about that, but potential Xcode11+ iOS13 clients might need it to solve similar issues we had before
> > 
> > 
> > I find this "support" for Xcode11+ / iOS 13 to be... weird. We're not actually checking that everything works with those versions (manually or in CI) and I have a feeling the storybook target will not work for precisely the reason that we have those static/dynamic targets.
> 
> Agree, let's remove it for now. We can always easily re-add it, if anybody would truly need it

Done separately in #348.

PavelHolec (approved):

Issue #241, opened 2022-07-22
Add DynamicTypeSize support to Timeline


Parabak:
In case issue #195 will be merged, main issue to fully support DynamicTypeSize  will be to find where exactly the icon is located.  
The icon might be aligned either to a first text line or to the container top.
Icon size is a dynamic value, while in timeline we use a constant

Also we should expose init with **both custom header and content** so user will be more flexible with desired design.  As an example, now horisontal padding is hardcoded which leads to a bigger gap, when  custom header is used instead of badge component.

PavelHolec:
Basic fixes were done as part of #315

PR #346, opened 2022-10-28
Update timeline design

bul-nick-al:
Snímek obrazovky 2022-10-26 v 15 13 29

Update the `Timeline` component to the latest design.
1. Add animation for the current state 
2. Extend `TimelineItemStyle` to support past, current and future items
3. Make the progress line dashed for future / inactive steps
4. Make progress line's color a gradient of colors of the steps it connects
5. Replace `Badge` with `Heading` in `TimelineSteps`'s Header
6. Remove `TimelineStepBadgeText`, `TimelineStepBottomText`, `Content` and `Header` from `TimelineItem`
7. Rename `TimelineStep` to `TimelineItem`

Animations: 

https://user-images.githubusercontent.com/13184609/198676302-772473e9-7dd4-461f-9ec7-15ef3dc1baf8.mov


Sources/Orbit/Support/Components/TimelineStep.swift
```diff
@@ -52,49 +57,82 @@ public struct TimelineStep<Header: View, Content: View>: View {
         .anchorPreference(key: TimelineStepPreferenceKey.self, value: .bounds) {
             [TimelineStepPreference(bounds: $0, style: style)]
         }
+        .onAppear { shouldAnimate = !reduceMotion }
```
PavelHolec, line 60: As most usages are in ternary operators, we might not need the additional @State property and `reduceMotion` can be used directly.

Sources/Orbit/Components/Timeline.swift
```diff
@@ -77,31 +80,31 @@ struct TimelinePreviews: PreviewProvider {
             TimelineStep(
                 "Booked",
                 sublabel: "January 3, 10:43",
-                style: .status(.success),
+                style: .pastSuccess,
```
PavelHolec, line 83: For the snapshot, one of the steps should be "active" for completeness (we probably have some more exhaustive in the previews already that can be used for snapshot)

PavelHolec:
👏 nice!

Sources/Orbit/Components/Timeline.swift
```diff
@@ -77,31 +80,31 @@ struct TimelinePreviews: PreviewProvider {
...
                 description: "Be at your departure gate at least 30 minutes before boarding."
             )
             TimelineStep(
                 "Board",
                 sublabel: "May 4, 8:15",
-                style: .status(.critical),
+                style: .pastCritical,
                 description: "Be at your departure gate at least 30 minutes before boarding."
             )
             contentPlaceholder
```
PavelHolec, line 104: Preview should be updated to use the custom content directly, without client having to use `TimelineStepPreference` (this type should probably not be public at all)

```
            TimelineStep(
                "Board",
                sublabel: "May 4, 8:15",
                style: .pastCritical,
                description: "Be at your departure gate at least 30 minutes before boarding."
            ) {
                contentPlaceholder
            }
```

Inits might need to be updated, as this combination seems to be missing. Ideally, `content` should be renamed to `footer` so that call site can use full combination of header + default content + footer.

Sources/Orbit/Support/Components/TimelineStep.swift
```diff
@@ -24,12 +24,17 @@ public extension VerticalAlignment {
...
+    @Environment(\.accessibilityReduceMotion) var reduceMotion
+
+    let animationSpeed: CGFloat = 0.25
 
     let style: TimelineStepStyle
     let isIconFirstTextLineCentered: Bool
     @ViewBuilder let header: Header
     @ViewBuilder let content: Content
 
+    @State var shouldAnimate = false
```
PavelHolec, line 36: The animation does not seem to be running under live preview canvas

Sources/Orbit/Support/Components/TimelineStep.swift
```diff
@@ -147,32 +185,59 @@ public extension TimelineStep where Header == TimelineStepBadgeText, Content ==
 // MARK: - Types
 
 public enum TimelineStepStyle: Equatable {
-    case `default`
-    case status(Status)
+    case inactive
```
PavelHolec, line 188: State/variant names should be taken 1:1 from design when possible, to avoid creating arbitrary names.

Parabak:
The issue with `DynamicTypeSize` suppoert https://github.com/kiwicom/orbit-swiftui/issues/241 is still relevant, right?

PavelHolec:
> The issue with `DynamicTypeSize` support #241 is still relevant, right?

From what I quickly checked, we might close it - this update seems to cover that implementation as well.

Sources/Orbit/Components/Timeline.swift
bul-nick-al: I am not sure, if it should be done this way. At the moment there is no case in design where we would need to place custom content in-between the header and the footer. In fact, there is no footer, there is rather a description text, so maybe header + description + content would be a better idea? But honestly, I think that keeping it the way it's done now, with a public `TimelineStepPreference` is better because the content passed to the timeline is not type restricted so the client can pass other views there anyways. And at the moment they have a mechanism to draw the timeline correctly even if they decide to shove something else in there. If we make it internal, any custom view will break the layout and there won't be any mechanism to fix it. What do you think?

Sources/Orbit/Support/Components/TimelineStep.swift
bul-nick-al: in those ternary operators it is crucial that the variable changes its state from `false` to `true` in order to launch animation. If we use `reduceMotion` directly, the animation does not start

PavelHolec: I got it now. Then maybe lets rename it to `animationLoopTrigger` or similar, because this sounds like a simple property of whether the component should animate or not.

Sources/Orbit/Components/Timeline.swift
PavelHolec: `header + description + content` sure, it is just that footer would be more explicit in where the content would be put, as there are multiple of them now.

> if they decide to shove something else in there

Sure, but this something will always be a part of some step. I would not put this burden on caller - the simplier the API is, the better.

Sources/Orbit/Components/Timeline.swift
bul-nick-al: I added the active state and face some weird layout in the snapshots (you can check in the diffs). Do you know what might be causing it or how it can be fixed? In the previews all looks fine
image

And another question: can we more to Orbit `IsPreviewLayoutScreenContentKey`? Or is there something similar already? I'd like to have more control over animation state if it's for a snapshot

Sources/Orbit/Components/Timeline.swift
bul-nick-al: Okay, added `Footer` and another initialiser

Sources/Orbit/Support/Components/TimelineStep.swift
bul-nick-al: renamed

Sources/Orbit/Components/Timeline.swift
PavelHolec: I will check the layout. For the `IsPreviewLayoutScreenContentKey`, it might be ported in future, but would it alone help? Here, we just need to decide what animation state we want to have as "starting" one (non-animated version) and loop the animation around that point. (snapshots test code can simply force reduceMotion environment to false in order to have the state non-animated)

Sources/Orbit/Support/Components/TimelineSegmentLine.swift
```diff
@@ -0,0 +1,35 @@
+import SwiftUI
+
+struct TimelineSegmentLine: View {
```
PavelHolec, line 3: Is the term TimelineSegmentLine driven by an existing design? In designs, I only found `TimelineLine` component. Those design atomic parts were introduced recently, so even our existing `TimelineStep` is not accurate now and might be renamed in future, but this component seems to already have a name.

Also, as it is internal only, it should not appear in the Timeline.md file

Sources/Orbit/Support/Components/TimelineStep.swift
```diff
@@ -47,61 +53,97 @@ public struct TimelineStep<Header: View, Content: View>: View {
...
+public extension TimelineStep where Header == TimelineStepHeadingText {
 
     /// Creates Orbit TimelineStep component with custom details.
     init(
         _ label: String,
         sublabel: String = "",
-        style: TimelineStepStyle = .default,
-        @ViewBuilder content: () -> Content
+        style: TimelineStepStyle = .inactive,
+        @ViewBuilder content: () -> Content,
```
PavelHolec, line 131: I thought we could only rename the existing `content` -> `footer`, not requiring to support both at the same time. But maybe we do need both? 🤔

Similar to components like ListChoice, call site can leave some fields empty and use custom content or it can use both. Can similar api work here as well?

Sources/Orbit/Components/Timeline.swift
```diff
@@ -20,12 +20,15 @@ public struct Timeline<Content: View>: View {
...
-                                .frame(
-                                    width: TimelineStepStyle.indicatorDiameter * sizeCategory.ratio,
-                                    height: progressLineHeight(for: index - 1, in: preferences, geometry: geometry)
-                                )
+                            TimelineSegmentLine(
+                                height: progressLineHeight(for: index - 1, in: preferences, geometry: geometry),
+                                startPointStyle: preferences[index - 1].style,
+                                endPointStyle: preferences[index].style
+                            )
+                            .frame(
```
PavelHolec, line 28: The `frame()` could be encapsulated inside the line component.

Sources/Orbit/Support/Components/TimelineStep.swift
```diff
@@ -21,14 +21,20 @@ public extension VerticalAlignment {
...
+public struct TimelineStep<Header: View, Content: View, Footer: View>: View {
 
     @Environment(\.sizeCategory) var sizeCategory
+    @Environment(\.accessibilityReduceMotion) var reduceMotion
+
+    let animationSpeed: CGFloat = 0.25
 
     let style: TimelineStepStyle
     let isIconFirstTextLineCentered: Bool
     @ViewBuilder let header: Header
```
PavelHolec, line 33: With the new design, maybe we do not need a customizable header anymore and the only customizable content could be the footer.

Sources/Orbit/Components/Timeline.swift
PavelHolec: For the snapshots, this is probably animation issue - try moving the frame modifier after the animation modifier.

Sources/Orbit/Support/Components/TimelineStep.swift
PavelHolec: The enum will not prevent caller to use steps in wrong order or impossible combinations. We can keep the "status" property for each step (based on design - "default", "active", "success",...) and the information about which step is active can either be done similar to how the Tags work (preferred) - with an `index` parameter of the whole Timeline (this makes sure only one step is "current" and makes it easier to map this information from client), or it could be a second parameter (isActive) of each step.

PavelHolec:
The `TimelineStepHeadingText.content` can be updated so it checks whether both label and sublabel are empty.

Sources/Orbit/Support/Components/TimelineStep.swift
bul-nick-al: Removed

Sources/Orbit/Components/Timeline.swift
bul-nick-al: I replaced it with `offset`, not it will be more clear why it is used outside the component

Sources/Orbit/Support/Components/TimelineStep.swift
bul-nick-al: Left only `footer`

Sources/Orbit/Support/Components/TimelineSegmentLine.swift
bul-nick-al: Renamed

Sources/Orbit/Support/Components/TimelineStep.swift
bul-nick-al: Changed cases to `past, current(Status), future`

Sources/Orbit/Components/Timeline.swift
PavelHolec: The state should be done so that both devices with reduced motion and snapshots show the same thing.

It seems to look fine for `Timeline` component, but not for `TimelineItem` alone - not an issue though, as long as the `Timeline` itself works fine.

Sources/Orbit/Support/Components/TimelineStep.swift
PavelHolec: Nice. Now the `TimelineItem where Footer == EmptyView` init should reuse the main init.

Sources/Orbit/Support/Components/TimelineItem.swift
```diff
@@ -0,0 +1,440 @@
...
+}
+
+// MARK: - Types
+
+public enum TimelineItemStyle: Equatable {
+    case past
+    case current(Status)
+    case future
+
+    public static let indicatorDiameter: CGFloat = Icon.Size.large.value
```
PavelHolec, line 218: This probably belongs more to the TimelineItem component

Sources/Orbit/Support/Components/TimelineItem.swift
```diff
@@ -0,0 +1,440 @@
...
+    public typealias Value = [TimelineItemPreference]
+
+    public static var defaultValue: Value = []
+
+    public static func reduce(value: inout Value, nextValue: () -> Value) {
+        value.append(contentsOf: nextValue())
+    }
+}
+
+public struct TimelineItemPreference {
```
PavelHolec, line 276: Can be now removed?

PavelHolec, line 276: Can this and the key stay internal?

Sources/Orbit/Support/Components/TimelineItem.swift
```diff
@@ -0,0 +1,440 @@
...
+
+                Button("Check in")
+            }
+        }
+        .padding()
+        .previewDisplayName("Snapshots")
+    }
+}
+
+struct TimelineItemDynamicTypePreviews: PreviewProvider {
```
PavelHolec, line 361: We can now remove the dynamic previews

Sources/Orbit/Support/Components/TimelineItem.swift
bul-nick-al: This property is used both in `Timeline` and `TimelineItem`, at the same time it cannot be a static property of `TimelineItem` because it has a generic type. Doesn't it make sense to keep it in here?

Sources/Orbit/Support/Components/TimelineItem.swift
bul-nick-al: Now it can but I am not sure if it should. I think it would be nice to leave to clients the ability to insert their custom views into the hierarchy of `Timeline` without breaking the layout. One could argue that now it can be placed inside the footer, but let's image that, for whatever reason, a different approach will be needed. When it happened last time, I had to copy the entire component because such functionality was not supported. Client won't have to use this preference anymore, 99% chance that the footer will be enough. But why not to keep it just in case?

Sources/Orbit/Components/Timeline.swift
bul-nick-al: The animation is fixed by adding what value to observe for animation

Sources/Orbit/Support/Components/TimelineItem.swift
```diff
@@ -0,0 +1,413 @@
+import SwiftUI
+
+public extension VerticalAlignment {
```
PavelHolec, line 3: Can be internal and moved below after view definition, near typ extensions (example: `List` component, there we actually want the alignment to be available to align things below horizontally)

Sources/Orbit/Support/Components/TimelineItem.swift
```diff
@@ -0,0 +1,413 @@
...
+    @Environment(\.horizontalSizeClass) var horisontalSizeClass
+    @Environment(\.accessibilityReduceMotion) var reduceMotion
+
+    let label: String
+    let sublabel: String
+    let description: String
+    let style: TimelineItemStyle
+    @ViewBuilder let footer: Footer
+
+    var hasHeaderContent: Bool {
```
PavelHolec, line 36: These properties should go below viewbuilder properties

Sources/Orbit/Support/Components/TimelineItem.swift
```diff
@@ -0,0 +1,413 @@
...
+    }
+
+    var animation: Animation {
+        Animation.easeInOut.repeatForever().speed(0.25)
+    }
+
+    @State var animationLoopTrigger = false
+
+    public var body: some View {
+        HStack(alignment: .timelineItemAlignment, spacing: .small) {
```
PavelHolec, line 51: We should reuse the `firstTextBaseline` alignment guide that we use for these. We can just compute the firstTextBaseline for the indicator.

Sources/Orbit/Support/Components/TimelineItem.swift
PavelHolec: But it is a preference of the TimelineItem specifically, how else can client use it? 🤔 Let's start with what Orbit defines and only later unlock advanced features, if it is needed.

Sources/Orbit/Support/Components/TimelineItem.swift
bul-nick-al: Good point, removed

Sources/Orbit/Support/Components/TimelineItem.swift
```diff
@@ -0,0 +1,302 @@
...
+    let sublabel: String
+    let description: String
+    let type: TimelineItemType
+    @ViewBuilder let footer: Footer
+
+    public var body: some View {
+        HStack(alignment: (hasHeaderContent || hasDescription) ? .firstTextBaseline : .top, spacing: .small) {
+
+            TimelineIndicator(type: type)
+                .frame(
```
PavelHolec, line 24: The `frame` and `alignmentGuide` can be encapsulated inside the TimelineIndicator (it seems it partially already is, at least the frame is already inside)

bul-nick-al: Moved inside the component

Sources/Orbit/Support/Components/TimelineItem.swift
```diff
@@ -0,0 +1,302 @@
...
+    func iconContentBaselineOffset(height: CGFloat) -> CGFloat {
+        Icon.Size.small.textBaselineAlignmentGuide(sizeCategory: sizeCategory, height: height)
+    }
+}
+
+// MARK: - Inits
+
+public extension TimelineItem {
+
+    /// Creates Orbit TimelineItem component with text details.
```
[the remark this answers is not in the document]
bul-nick-al: Added

PavelHolec (approved):

Sources/Orbit/Support/Components/TimelineItem.swift
```diff
@@ -0,0 +1,291 @@
...
+            TimelineItem(
+                "Requested",
+                sublabel: "3rd May 14:04",
+                type: .future,
+                description: "We’ve assigned your request to one of our agents."
+            )
+            TimelineItem(
+                "Requested",
+                sublabel: "3rd May 14:04",
+                type: .current(.info),
```
hrach, line 204: IMO it is not natural to mark the "current"/"in progress" step with info style - info style looks blue in the whole Orbit. Reusing it here doesn't make user confident what to use.

PavelHolec: If needed, we can create specific `TimelineIndicator.Status` type that will be limited compared to normal `Status` (the status is mostly a property of the indicator/line anyway)

Sources/Orbit/Support/Components/TimelineIndicator.swift
```diff
@@ -0,0 +1,89 @@
...
+        }
+    }
+
+    var animation: Animation {
+        Animation.easeInOut.repeatForever().speed(0.25)
+    }
+
+    func iconContentBaselineOffset(height: CGFloat) -> CGFloat {
+        Icon.Size.small.textBaselineAlignmentGuide(sizeCategory: sizeCategory, height: height)
+    }
```
PavelHolec, line 88: If we will need to do Status-related adjustments anyway, it would be nice to have a preview for this component as well (to see all those possible states)

Sources/Orbit/Support/Components/TimelineItem.swift
bul-nick-al: In Figma they use 3 out of 4 Status states and then default and active

image


image

What would you propose in this case? 
1. Extend the `Type` enum and just ignore `.current(.info)`:
```swift
enum TimelineItemType  {
    case past
    case status(Status)
    case active
    case future
}
```
2. Untie from `Status` completely:
```swift
enum TimelineItemType  {
    case past
    case currentWarning
    case currentSuccess
    case currentCritical
    case currentActive
    case future
}
```

Or something else?

Sources/Orbit/Support/Components/TimelineIndicator.swift
bul-nick-al: I wouldn't do it for the same reason why I wouldn't do it for `TimelineLine` – they are both `internal` and their states are still tested anyways via snapshots of `Timeline` and `TimelineItem`. In `TimelineItem` you can see all the possible states of the indicator already. what do you think?

Sources/Orbit/Support/Components/TimelineItem.swift
PavelHolec: If i did not forget some state, maybe it could be:
 

```
extension TimelineIndicator {

   enum Progress {
      case past          // Always 'done'
      case present(Status? = nil)
      case future      // Always 'grey'
  }

   enum Status {
      case warning
      case critical
  }
}
```

bul-nick-al: Updated, although kept it as `TimelineItemType` to maintain the naming as in FIgma

PR #347, opened 2022-10-29
Check types are referenced in DocC

sjavora:
This is a plugin that checks whether all top-level public types are mentioned in the DocC bundle. This should prevent the generated documentation from showing undocumented types on the main page. This only works for types, not (free) functions!


PavelHolec (approved):
Nice 👍

sjavora:
Of course, this won't pass until somebody writes something about `TextRepresentable` 😉 

Screenshot 2022-10-29 at 17 22 15

PavelHolec:
We could update the comment to this for now

```
/// A type representable as `SwiftUI.Text`.
```

Issue #173, opened 2022-06-13
Use native SwiftUI environment keys


PavelHolec:
Some components might use existing environment keys or modifiers in favour of parameters.

#### Available related environment keys

- #401
- #404 
  - `lineLimit`, `lineSpacing`, `multilineTextAlignment `, `textCase`, `truncationMode `: could be used in Text/Heading to be able to inject text-based modifiers inside.
  - A `font` override could be supported as well.
- `defaultMinListRowHeight`
  - can be used to support ListChoice
- `accessibilityReduceMotion`
  - should be respected whenever we do animations
  - ~~can be also used in snapshot test code to force non-animated version~~ - This environment key is read-only 🤷 
- `tint`
  - can be used to override the accent color of some elements (`button`, `checkbox`, `switch`)

#### iOS14+ modifiers
- `isFocused` (iOS14+)
  - could be used whenever we have similar flag or enum
  - works correctly on `InputField` when used from iOS14+ client
- `controlSize` (iOS15+)
  - could be used for sizing Badges or Buttons (whenever we have size enum)

#### Related modifiers
- `foregroundColor()` and `tint()` (iOS15+): 
  - write-only modifier 🤷 Can be partially [accessed](https://kylehugh.es/documents/how-to-read-foreground-and-tint-colors-in-swiftui-views/) by:
    - `ForegroundShapeStyle()`
    - `TintShapeStyle()`
    - But we cannot access the underlying UIKit color that we need for the TextLink overlay or other formatting implementation

sjavora:
https://www.fivestars.blog/articles/swiftui-environment-values/

PR #351, opened 2022-11-01
Support text formatting for `Heading`

PavelHolec:
Removed the forced uppercasing of lvl6 title (it would not work together with formatting anyway).

image

Concatenated

image

### DIFF
--- a/Sources/Orbit/Components/Heading.swift
+++ b/Sources/Orbit/Components/Heading.swift
@@ -8,13 +8,15 @@ public struct Heading: View {
 
     @Environment(\.sizeCategory) var sizeCategory
 
-    let label: String
+    let content: String
     let style: Style
     let color: Color?
+    let lineSpacing: CGFloat?
     let alignment: TextAlignment
+    let accentColor: UIColor
 
     public var body: some View {
-        if updatedLabel.isEmpty == false {
+        if content.isEmpty == false {
             text(sizeCategory: sizeCategory)
                 .multilineTextAlignment(alignment)
                 .fixedSize(horizontal: false, vertical: true)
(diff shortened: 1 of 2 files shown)


Sources/Orbit/Components/Heading.swift
```diff
@@ -23,54 +25,77 @@ public struct Heading: View {
...
 
 // MARK: - Inits
 public extension Heading {
 
     /// Creates Orbit Heading component.
+    ///
+    /// - Parameters:
+    ///   - content: String to display. Supports html formatting tags `<strong>`, `<u>`, `<ref>`.
+    ///   - style: Heading style.
+    ///   - color: Font color. Can be set to `nil` and specified later using `.foregroundColor()` modifier.
```
sjavora, line 50: Hmm why do we have the parameter then? Do you want to switch to the "native" way of doing it everywhere at once?

PavelHolec: It is a tricky thing that I did not yet found a better solution for. While it seems it could be solved by !173, the main current issue is that our Text and Heading have a default color set to a specific Orbit color (inkDark), so that a caller does not need to remember specifying it every time with a `foregroundColor` modifier. 

Problem is, what I believe is a SwiftUI bug, that whenever `foregroundColor` is already applied once (inside the component), it cannot be overridden further (at a call site). So we cannot easily say "use default inkDark, but then let call site override it". That's why the current "duplicate" approach exists.

sjavora: I'd add that explanation to #173, if it's not already mentioned somewhere else.

sjavora (approved):

PR #356, opened 2022-11-02
Make `AccessibilityID` reusable and extensible by client

### LINKED ISSUE #301: Should `Select` use `.accessibility(.inputValue)`?
While writing some page objects for UI tests, I noticed that `Select` uses `.accessibility(.inputValue)`... but `Select` is not an input... is this a copy/paste error is it intended? I'd create `.selectValue` rather than reuse `.inputValue`...

cc @PavelHolec

PavelHolec:
Should also fix #301


Sources/Orbit/Support/Accessibility/AccessibilityID.swift
```diff
@@ -1,56 +1,72 @@
...
+    static let tileDisclosureIcon: Self             = "orbit.tile.disclosure.icon"
+
+    public let rawValue: RawValue
+
+    public init(rawValue: String) {
+        self.rawValue = rawValue
+    }
+}
+
+extension AccessibilityID: ExpressibleByStringLiteral {
```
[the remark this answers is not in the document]
PavelHolec: Agree, reverted. 

Also introduced the iOS14 preferred `accessibilityModifier` parameter.

Sources/Orbit/Support/Accessibility/AccessibilityID.swift
```diff
@@ -1,57 +1,73 @@
...
-    case tileTitle                      = "orbit.tile.title"
-    case tileIcon                       = "orbit.tile.icon"
-    case tileDescription                = "orbit.tile.description"
-    case tileDisclosureButtonLink       = "orbit.tile.disclosure.buttonlink"
-    case tileDisclosureIcon             = "orbit.tile.disclosure.icon"
+///
+/// Can be extended to provide custom accessibility identifiers.
+public struct AccessibilityID: RawRepresentable {
+
+    static let alertButtonPrimary           = Self(rawValue: "orbit.alert.button.primary")
```
sjavora, line 8: What about moving the acutal definitions to an extension? That way, it will be the same as in clients and the primary declaration can focus on `RawRepresentable` only.

Sources/Orbit/Support/Accessibility/AccessibilityID.swift
```diff
@@ -1,57 +1,73 @@
...
+    @available(iOS 14.0, *)
+    func accessibilityIdentifier(_ accessibilityID: AccessibilityID) -> some View {
+        self.accessibilityIdentifier(accessibilityID.rawValue)
+    }
 
-    /// Uses the specified Orbit identifier to identify the view inside a component.
-    func accessibility(_ identifier: AccessibilityID) -> some View {
-        self.accessibility(identifier: identifier.rawValue)
+    /// Uses the specified identifier to identify the view inside a component.
+    @available(iOS, introduced: 13.0, deprecated: 100000.0, renamed: "accessibilityIdentifier(_:)")
```
sjavora, line 69: Shouldn't that be `deprecated: 14.0`? 🤔

Sources/Orbit/Support/Accessibility/AccessibilityID.swift
PavelHolec: Good idea - and it can probably go as extension in each component file separately, right?

PavelHolec: Like so?

image

sjavora: Why not, although are there no cases where an identifier is used in multiple files?

Sources/Orbit/Support/Accessibility/AccessibilityID.swift
PavelHolec: I copied the SwiftUI declaration, but it is true that we now specify it exactly. According to docs, it should be 16.1, right?

https://developer.apple.com/documentation/swiftui/presentedwindowcontent/accessibility(identifier:)

Sources/Orbit/Support/Accessibility/AccessibilityID.swift
PavelHolec: Did not happen so far. Even if yes, it would be defined in the component it primarily belongs to.

sjavora (approved):
Nice.

PR #362, opened 2022-11-08
Configure the entire timeline with one index and style

bul-nick-al:


hrach:
I was thinking about this lot and I probably don't want to go this path. I'm posting here just for FYI. My reasoning:

We want to utilize composition in a way that the item/step is independent from its parent container. That allows you to delegate a lot of functionality and resolution to the logic that is handling the particular step rendering. 

If the step would not be "independent", i.e. (part of) its visual would be decided/affected by a parent container, then there is a question, why not make the component a single call completely, e.g. something like this in Kotlin (simplified):

```kotlin
Timeline(
   steps = listOf("Step 1", "Step 2"),
   activeStep = 1,
)
```

Though, I'm still not convinced about the composition approach.

PavelHolec:
> I was thinking about this lot and I probably don't want to go this path. I'm posting here just for FYI. My reasoning:
> 
> We want to utilize composition in a way that the item/step is independent from its parent container. That allows you to delegate a lot of functionality and resolution to the logic that is handling the particular step rendering.
> 
> If the step would not be "independent", i.e. (part of) its visual would be decided/affected by a parent container, then there is a question, why not make the component a single call completely, e.g. something like this in Kotlin (simplified):
> 
> ```kotlin
> Timeline(
>    steps = listOf("Step 1", "Step 2"),
>    activeStep = 1,
> )
> ```
> 
> Though, I'm still not convinced about the composition approach.

The `TimeLine` **is** a single call. Whether the data for steps are provided using a list parameter or using the @ViewBuilder parameter depends on further usage. The @ViewBuilder may be more composable and flexible, especially when we want to allow caller to provide custom @ViewBuilder closure for each step.

We can also keep the current approach where it is callers responsibility to model state correctly (after all, it is possible that there will be multiple active states or more combinations of past/future+status states). 

This draft goal is now mostly to:
1) Provide similar interface to `Tabs` (which has similar semantics - only one item can be active at a time)
2) Proof of concept for any future similar composited components

PavelHolec:
Closing as obsolete for now.

PR #378, opened 2022-11-30
Include more support files

PavelHolec:
Moved some updated code from Kiwi.com


Sources/Orbit/Components/Switch.swift
```diff
@@ -117,9 +117,7 @@ struct SwitchPreviews: PreviewProvider {
     static var previews: some View {
         PreviewWrapper {
             standalone
-            storybook
-            storybook
-                .preferredColorScheme(.dark)
```
sjavora, line 122: Noticed this in a couple of files - don't we want the dark mode previews anymore?

Sources/Orbit/Support/Components/BarButton.swift
```diff
@@ -3,36 +3,66 @@ import SwiftUI
...
+            } else {
+                foregroundColor = .inkDark
+            }
+
+            return configuration.isPressed ? foregroundColor.opacity(0.6) : foregroundColor
+        }
+    }
+}
+
+extension SwiftUI.ButtonStyle where Self == BarButton.ButtonStyle {
```
sjavora, line 51: Shouldn't this be public?

Sources/Orbit/Support/Previews/PreviewDisplayName.swift
```diff
@@ -0,0 +1,23 @@
+import SwiftUI
+
+public extension View {
+
+    /// Sets a name to show in the canvas for a preview, based on a function name.
```
sjavora, line 5:
```suggestion
    /// Sets a name to show in the canvas for a preview, based on the preview property name.
```

Sources/Orbit/Components/Switch.swift
PavelHolec: It is the same case as with dynamic size - the Xcode 14 preview support makes this obsolete, as we can see any variant combination with few clicks.

Sources/Orbit/Support/Components/BarButton.swift
PavelHolec: I thought about keeping it internal, so that the primary way to use this component would be through the `BarButton` view itself, in order not to pollute the SwiftUI.ButtonStyle namespace to call site.

Sources/Orbit/Support/Previews/PreviewDisplayName.swift
PavelHolec: Will update. Its just that the internal parameter is called `function` - should I rename it to be `property`, as we expect to use it on properties only?

sjavora: Well, we expect it to be used on properties, since the previews are computed `var`s...

PavelHolec: Yup, updated

sjavora (approved):

username0x0a (approved):

PR #379, opened 2022-11-30
Update BarButton and NavigationButton

PavelHolec:
Update layout of BarButton and NavigationButton components. Added snapshot tests.


Sources/Orbit/Support/Components/BarButton.swift
```diff
@@ -1,9 +1,10 @@
 import SwiftUI
 
-/// Button for NavigationBar.
+/// An icon-based bar button for suitable for toolbar actions inside `NavigationBar`.
```
sjavora, line 3: Is there such a thing as `NavigationBar` in SwiftUI?. It's the `.toolbar` modifier, right? `.navigationBar*` should probably not be used in iOS 14+?

PavelHolec: For iOS14+ the `toolbar` will be the way to go. 

As for now, the iOS13 modifier we have to use is still called `navigationBar*` (`navigationBarItems`). So maybe more correct would be to remove the toolbar mention or mention it with iOS14 note.

PavelHolec: The Apple API specifies `toolbar` as `Populates the toolbar or navigation bar with the views you provide. `, so we can use similar `toolbar or navigation bar` naming

sjavora: Sure, half the issue I have is the capitalization... `NavigationBar` sounds like some concrete type, which is not the case. `toolbar or navigation bar` sounds good to me.

sjavora (approved):

PR #391, opened 2022-12-08
Automatically update snapshots in relevant branches

### LINKED ISSUE #286: Automatically update snapshots in relevant branches
If tokens change, it makes sense that snapshots will need to change as well.

We might also want to think about regenerating snapshots in all merge requests by default 🤔

sjavora:
I'm sure this could be improved by somehow cleaning up existing snapshot update commits, but this should still be useful on its own.

Resolves #286.


PavelHolec (approved):

sjavora:
I had another idea - run regeneration only if the test jobs fail - it should minimize redundant jobs. I'll try setting that up today..

PR #397, opened 2022-12-19
Use fixed frame for `TextStrut`

PavelHolec:
With this change, the `TextStrut` (a component that maintains some minimal text-size driven height for other components) gets a consistent size, independent on currently used font. A benefit is that the components are now more stable in size and match the Orbit definition (for the default, large, dynamic type).

Side effect is that this will require most snapshots to be updated.

image


PavelHolec:
Sizes need to be rechecked. Using `CircularPro` should match orbit heights in normal font sizes.

Sources/Orbit/Components/Badge.swift
```diff
@@ -8,7 +8,7 @@ import SwiftUI
 /// - Note: [Orbit definition](https://orbit.kiwi/components/badge/)
 public struct Badge: View {
 
-    public static let verticalPadding: CGFloat = 4 + 1/3 // Results in ±24 height at normal text size
+    public static let verticalPadding: CGFloat = 4 // Results in ±24 height at normal text size
```
sjavora, line 11: `.xxSmall`?

sjavora (approved):

Sources/Orbit/Components/Badge.swift
PavelHolec: Fixed 👍

sjavora (approved):

PavelHolec:
Will be replaced by `Strut` with simpler calculation

PR #398, opened 2022-12-19
Add `Collapse` component

### LINKED ISSUE #352: Create `Collapse` component
image

sjavora:
Should more things be customizable? Mainly thinking of the title/header and whether the separator should be hideable...

Resolves #352.

testCollapse 1

https://user-images.githubusercontent.com/12349477/208420408-13b07028-09d6-451c-b93b-eed4369fb6c1.mov


PavelHolec:
The version with fixed `Text` is good enough for what Orbit provides in design. In real usage, the non-collapsed content will be more custom than simple text though.

On top of the existing `String` based init, the component should accept two ViewBuilder arguments - the collapsed version (`header`) and the expanded `content` (these two will be concatenated in the VStack). The String-based init will then call this more generic one.

The header will probably again use `HStack` with custom content, spacer and trailing chevron.

sjavora:
Ok, what about some highlight on touchdown for the header? Currently there is none, but maybe there should be something similar to `Tile`?

sjavora:
@PavelHolec updated with header content.

PavelHolec:
> Ok, what about some highlight on touchdown for the header? Currently there is none, but maybe there should be something similar to `Tile`?

That would be nice. The highlight could be done as a background behind the header, driven by the Button `isPressed` state. This would expect the header to have a transparent background though. It would be worth deciding how to unify this behaviour with the similar components like Tile/ListChoice in a follow-up.

Sources/Orbit/Components/Collapse.swift
```diff
@@ -0,0 +1,129 @@
...
+                }
+            } label: {
+                HStack(spacing: 0) {
+                    header
+                        .accessibility(.collapseHeader)
+                    Spacer(minLength: .xSmall)
+                    Icon(.chevronDown)
+                        .rotationEffect(isExpanded ? .degrees(180) : .zero)
+                }
+                .padding(.vertical, .small)
```
PavelHolec, line 35: This padding might be too specific. Maybe it could be used only when constructing the `String`-based simple version. That way, a call site that provides custom header content would be fully in charge of any required padding.

A `Card` component embeds some "typical" padding configurations as a part of its layout configuration. For this component, that solution might not be necessary.

Sources/Orbit/Components/Collapse.swift
```diff
@@ -0,0 +1,129 @@
...
+        .overlay(Separator(), alignment: .bottom)
+    }
+}
+
+public extension Collapse {
+
+    /// Creates Orbit Collapse component.
+    init(isExpanded: Bool = false, @ViewBuilder content: () -> Content, @ViewBuilder headerContent: () -> Header) {
+        self.header = headerContent()
+        _isExpanded = .init(initialValue: isExpanded)
```
PavelHolec, line 53: If there is a chance that a call site would like to be informed about the state change, we would probably switch to a @Binding in future. We currently mix multiple approaches in our API already - a future follow-up might clean-up those that are less easy to work with. 

I expect that in future, we might replace some of our @State properties for @Bindings (or properties + callbacks) as we often later find out that a call site needs to be able to bind that state with its own mutable single source of truth.

Sources/Orbit/Support/Previews/PreviewProvider+Extensions.swift
```diff
@@ -1,6 +1,13 @@
 import SwiftUI
 
 extension PreviewProvider {
+
+    @ViewBuilder static var headerPlaceholder: some View {
```
PavelHolec, line 5: Handy 👍

Sources/Orbit/Components/Collapse.swift
sjavora: Whether view is expanded or not seems to fit the definition of `@State`... I wouldn't want to force users to pass in a `@Binding` in all circumstances - is there a way to get around that? If not, I'd probably go with a callback.

sjavora: Hmm... `DisclosureGroup` does have an init that takes a `Binding`... and also one that doesn't... I wonder how that works.

PavelHolec: I remember we did (or tried?) this already in some component - in the same way, there is an init both for a "normal" property and for binding. I believe two different internal properties are used under the hood at the worst case.

Another components may benefit from this double init (especially `Tag`). I will create a follow-up, but feel free to introduce the swiftui-like api in this story to see how it would work.

Sources/Orbit/Components/Collapse.swift
sjavora: Ok, set only for `Text` headers.

Sources/Orbit/Components/Collapse.swift
sjavora: Found [this article](https://www.fivestars.blog/articles/optional-binding/) which explores the same topic (and with a `DisclosureGroup` too, of course 😄) - used the ideas from it here.

Sources/Orbit/Components/Collapse.swift
```diff
@@ -0,0 +1,183 @@
...
+    /// Creates Orbit Collapse component with a binding to the expansion state.
+    init(isExpanded: Binding<Bool>, @ViewBuilder content: () -> Content, @ViewBuilder headerContent: () -> Header) {
+        self.header = headerContent()
+        self.headerVerticalPadding = 0
+        self.content = content()
+        self.isExpanded = isExpanded
+    }
+
+    /// Creates Orbit Collapse component.
+    init(@ViewBuilder content: () -> Content, @ViewBuilder headerContent: () -> Header) {
```
PavelHolec, line 68: We could still keep the `isExpanded: Bool = false` property in this init for a one-way direction.

sjavora: We could, but not even Apple does that for their components. I wouldn't overcomplicate it.

Sources/Orbit/Components/Collapse.swift
```diff
@@ -0,0 +1,183 @@
+import SwiftUI
+
+/// Hides long or complex information.
+///
+/// - Note: [Orbit definition](https://orbit.kiwi/components/collapse/)
+public struct Collapse<Header: View, Content: View>: View {
+
+    struct _Collapse: View {
```
PavelHolec, line 8: The article does not seem to need the internal structure. Do we have a specific need for it?

PavelHolec (approved):

PR #402, opened 2022-12-20
Replace `isEnabled` property with `isEnabled` environment key

PavelHolec:


Sources/Orbit/Components/ChoiceTile.swift
```diff
@@ -149,7 +149,7 @@ public struct ChoiceTile<Content: View>: View {
 
     @ViewBuilder var indicatorOverlay: some View {
         indicatorElement
-            .disabled(true)
+            .allowsHitTesting(false)
```
sjavora, line 152: What's the difference between this and the previous version?

Sources/Orbit/Components/InputField.swift
```diff
@@ -133,7 +134,7 @@ public struct InputField<Value>: View where Value: LosslessStringConvertible {
                 }
                 .accessibility(addTraits: .isButton)
                 .accessibility(.inputFieldPasswordToggle)
-                .opacity(state == .disabled ? 0 : 1)
+                .opacity(isEnabled ? 1 : 0)
```
sjavora, line 137: This won't ever happen if `isEnabled` is `false`.

Sources/Orbit/Support/Components/TagGroup.swift
```diff
@@ -23,8 +23,8 @@ public struct TagGroup<TM: TagModel>: View {
 
     @Environment(\.accessibilityReduceMotion) private var isReduceMotionEnabled
     @Environment(\.isFadeInEnabled) private var isFadeInEnabled
-    @State private var tagsFadeIn: [Bool] = []
-    @Binding private var tags: [TM]
+    @Binding var tags: [TM]
+    @State var tagsFadeIn: [Bool] = []
```
sjavora, line 27: Looks like we have an unwritten convention to order the property wrappers:
- `@Environment`
- `@Binding`
- `@State`

We should probably start a document with those, or at least an issue if we don't have enough of them yet.

Sources/Orbit/Components/ChoiceTile.swift
PavelHolec: Have not found yet a better way for these compositions. The scenario is that we want to display the indicator (radio/checkbox) **as if** it was enabled (disabling it here would cause it to be greyed out), yet we want it to not be accessible manually (accessibility itself is solved by the component, so the indicator alone should not be **interactive** by itself by any means).

Sources/Orbit/Components/InputField.swift
PavelHolec: Not sure I understand. We want to hide the eye icon toggle in case the field is disabled (the same condition as before, only reverted).

Sources/Orbit/Support/Components/TagGroup.swift
PavelHolec: Currently, it follows the loose order of how exposed the property is. 

It works for now, but maybe we may be breaking it in future, as the `@Binding` property is the one that can actually appear in the synthesized initializer for some internal helper views.

sjavora (approved):

Issue #77, opened 2022-03-25
Create `SegmentedSwitch` component


PavelHolec:
Can be probably created by mimicking or reusing parts of Tabs/Tab component or the native SegmentedControl.

https://www.figma.com/file/YMfJkmUMV8xiY7fgeW6QAr/%5BOrbit%5D-Native-Components?node-id=7798%3A3188

sjavora:
@Parabak this issue wasn't closed because you need to write exactly "Closes #77" or "Resolves #77" - those worked for me, anyway.

Closing this one manually.

PR #403, opened 2022-12-21
Implement SegmentedSwitch

Parabak:
Closing issue #77 

[Orbit details](https://orbit.kiwi/components/interaction/segmentedswitch/)


Sources/Orbit/Support/Components/Tab.swift
```diff
@@ -11,7 +11,7 @@ public struct Tab: View {
 
     public var body: some View {
         // FIXME: Convert to Button with .title4 style for a background touch feedback
-        HStack(spacing: 0) {
+        HStack(alignment: .center, spacing: 0) {
```
PavelHolec, line 14: This is a default value

Sources/Orbit/Components/SegmentedSwitch.swift
```diff
@@ -0,0 +1,157 @@
...
+        self.label = label
+    }
+}
+
+// MARK: - Previews
+struct SegmentedSwitchPreviews: PreviewProvider {
+    static var previews: some View {
+        PreviewWrapper {
+            segmentedSwitch()
+                .previewDisplayName("Unselected")
```
PavelHolec, line 124: It is better to extract each in a separate property (and choose one as a snapshot), as you will get a name for free with `previewDisplayName()`.

Sources/Orbit/Components/SegmentedSwitch.swift
```diff
@@ -0,0 +1,157 @@
...
+    let secondOption: Value
+
+    let separatorWidth: CGFloat = .xxxSmall
+    let borderWidth: CGFloat = .xxxSmall
+
+    public var body: some View {
+        FieldWrapper(
+            label,
+            message: message,
+            content: {
```
PavelHolec, line 24: content can be written as a trailing closure

Sources/Orbit/Components/SegmentedSwitch.swift
```diff
@@ -0,0 +1,157 @@
+import SwiftUI
+
+/// Showing two choices from which only one can be selected.
+///
+/// Currently SegmentedSwitch allows to have 2 segments.
+/// SegmentedSwitch can be in error state only in unselected state.
+///
+/// - Note: [Orbit definition](https://orbit.kiwi/components/interaction/segmentedswitch/)
+public struct SegmentedSwitch<Value>: View where Value: LosslessStringConvertible & Equatable {
```
sjavora, line 9: I don't remember such requirements on any other... "value" type. Is it really needed?

Sources/Orbit/Components/SegmentedSwitch.swift
```diff
@@ -0,0 +1,157 @@
...
+        RoundedRectangle(cornerRadius: BorderRadius.default)
+            .fill(backgroundColor)
+            .overlay(
+                RoundedRectangle(cornerRadius: BorderRadius.default)
+                    .stroke(borderColor, lineWidth: borderWidth)
+            )
+    }
+
+    private var message: Message? {
+        if case .error = _message, selection != nil {
```
sjavora, line 78: Would this code be longer as a `switch`? If not, please use that.

Sources/Orbit/Components/SegmentedSwitch.swift
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    private var borderColor: Color {
+        guard selection == nil else { return .clear }
+
+        switch message {
+            case .error:
+                return .redNormal
+            default:
+                return .cloudNormal
```
sjavora, line 96: Can be inlined

Sources/Orbit/Components/SegmentedSwitch.swift
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    private func computeSelectionOffset(in reader: GeometryProxy) -> CGSize {
+        .init(width: selection == firstOption ? 0 : reader.size.width / 2 - borderWidth, height: 0)
+    }
+
+    public init(
+        label: String,
+        message: Message? = nil,
+        firstOption: Value,
```
PavelHolec, line 107: We should follow existing APIs:
- `_ label: String,`
- `message` should go last, to follow the left-right, top-bottom order of components

```

    public init(
        _ label: String,
        firstOption: Value,
        secondOption: Value,
        message: Message? = nil,
        selectedOption: Binding<Value?>
    )
```

Sources/Orbit/Components/SegmentedSwitch.swift
```diff
@@ -0,0 +1,157 @@
...
+        secondOption: Value,
+        selection: Binding<Value?>
+    ) where Value: LosslessStringConvertible {
```
sjavora, line 110: Currently `Value` always has to be `LosslessStringConvertible` so this is redundant. I'd prefer it to be only here though.

PavelHolec, line 110: Is this necessary? Can we use `String` instead?

Sources/Orbit/Components/SegmentedSwitch.swift
```diff
@@ -0,0 +1,157 @@
...
+
+    public var body: some View {
+        FieldWrapper(
+            label,
+            message: message,
+            content: {
+                HStack(spacing: 0) {
+                    segment(title: firstOption)
+                    separator
+                    segment(title: secondOption)
```
PavelHolec, line 28: Each segment will need a new orbit AccessibilityIdentifier so a call site can reach both options

Sources/Orbit/Components/SegmentedSwitch.swift
```diff
@@ -0,0 +1,157 @@
...
+                }
+                .background(selectionBackground)
+                .background(roundedBackground)
+                .fixedSize(horizontal: false, vertical: true)
+            }
```
PavelHolec, line 33: The whole component should be accessible as a single component.

InputField example:

image

Sources/Orbit/Components/SegmentedSwitch.swift
sjavora: SwiftUI seems to use `Identifiable` rather than `Equatable` in similar cases.

Sources/Orbit/Components/SegmentedSwitch.swift
```diff
@@ -0,0 +1,157 @@
...
+        )
+    }
+
+    @ViewBuilder func segment(title: Value) -> some View {
+        Text(title.description)
+            .lineLimit(1)
```
PavelHolec, line 39: Should be removed to allow more verbose languages and dynamic type

Sources/Orbit/Components/SegmentedSwitch.swift
```diff
@@ -0,0 +1,157 @@
...
+            .padding(.vertical, .xSmall)
```
PavelHolec, line 40: `.padding(.xSmall)` to get horizontal padding as well

sjavora:
Besides fixing issues found by CI, be sure to add this to the Storybook as well.

Sources/Orbit/Components/SegmentedSwitch.swift
```diff
@@ -0,0 +1,157 @@
...
+            .frame(maxWidth: .infinity)
+            .contentShape(Rectangle())
+            .onTapGesture {
+                selection = title
+            }
+    }
+
+    @ViewBuilder var separator: some View {
+        Rectangle()
```
PavelHolec, line 49: Can be replaced with

```
borderColor
  .frame(width: .xxxSmall)
```

Sources/Orbit/Components/SegmentedSwitch.swift
```diff
@@ -0,0 +1,157 @@
...
+        switch message {
+            case .error:
+                return .redNormal
+            default:
+                return .cloudNormal
+        }
+    }
+
+    private func computeSelectionOffset(in reader: GeometryProxy) -> CGSize {
+        .init(width: selection == firstOption ? 0 : reader.size.width / 2 - borderWidth, height: 0)
```
PavelHolec, line 101: This calculation can return just CGFloat and be used as `.offset(x: this)`

Sources/Orbit/Components/SegmentedSwitch.swift
```diff
@@ -0,0 +1,157 @@
...
+public struct SegmentedSwitch<Value>: View where Value: LosslessStringConvertible & Equatable {
+
+    @Binding private var selection: Value?
+    let label: String
+    let _message: Message?
+    let firstOption: Value
+    let secondOption: Value
+
+    let separatorWidth: CGFloat = .xxxSmall
+    let borderWidth: CGFloat = .xxxSmall
```
PavelHolec, line 18: Should probably be equal to `BorderWidth.emphasis` to match other components

Sources/Orbit/Components/SegmentedSwitch.swift
```diff
@@ -0,0 +1,157 @@
...
+///
+/// Currently SegmentedSwitch allows to have 2 segments.
+/// SegmentedSwitch can be in error state only in unselected state.
+///
+/// - Note: [Orbit definition](https://orbit.kiwi/components/interaction/segmentedswitch/)
+public struct SegmentedSwitch<Value>: View where Value: LosslessStringConvertible & Equatable {
+
+    @Binding private var selection: Value?
+    let label: String
+    let _message: Message?
```
PavelHolec, line 13: `message`

Sources/Orbit/Support/Components/Tab.swift
Parabak: I have no idea where it comes from 😅

Sources/Orbit/Components/SegmentedSwitch.swift
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    @ViewBuilder var separator: some View {
+        Rectangle()
+            .fill(borderColor)
+            .frame(width: separatorWidth)
+    }
+
+    @ViewBuilder var selectionBackground: some View {
+        GeometryReader { reader in
```
PavelHolec, line 55: Could be replaced by Spacer version

```
HStack {
            if selection == secondOption {
                Spacer(minLength: 0)
                    .layoutPriority(1)
            }

            Color.whiteDarker
                .clipShape(RoundedRectangle(cornerRadius: BorderRadius.default))
                .padding(borderWidth)
                .elevation(.level1)
                .layoutPriority(1)

            if selection == firstOption {
                Spacer(minLength: 0)
                    .layoutPriority(1)
            }
        }
        .animation(.default, value: selection)
```

Sources/Orbit/Components/SegmentedSwitch.swift
Parabak: There is another property `private var message: Message? {` as we agreed to hide error message in case there is a selection.

So `let _message` is a value provided by client, `var message` is used in layout

Sources/Orbit/Components/SegmentedSwitch.swift
Parabak: If we leave it only in init, how we will pass `selection: Binding<Value?>`. My main idea was that if client uses e.g. `enum Gender` it's possible to pass `Binding<Gender?>`

Sources/Orbit/Components/SegmentedSwitch.swift
Parabak: `Equatable` is required in conditions like `if selection == secondOption` when component check what segment have to be highlighted. Another component using `LosslessStringConvertible` requirement is `InputField`

Sources/Orbit/Components/SegmentedSwitch.swift
Parabak: Alright, didn't know about this enum. Took 2pt value from Figma

Sources/Orbit/Components/SegmentedSwitch.swift
PavelHolec: Or, can we simply use the `message` from call-site, as is? We should show if client provides it

Sources/Orbit/Components/SegmentedSwitch.swift
PavelHolec: If there is no better idea, I would preffer the (not ideal either) `Tabs` approach, as this is a special case of `Tabs` anyway, so an ideal workaround is to make those APIs consistent.

Imagine that a call site is first fine with 2 options and then 3 options are needed - they would need to completely rewire the logic just because they needed to add a 3rd option.

Sources/Orbit/Components/SegmentedSwitch.swift
PavelHolec: We seem to have a mismatch there - will be fixed in a future follow-up.

Sources/Orbit/Components/SegmentedSwitch.swift
Parabak: we agreed that error msg can be used only when there is no selection.

Parabak:
Updated. Please have a look. 

I've failed to build Storybook though 🤷

Tests/SnapshotTests/Components/SegmentedSwitchTests.swift
```diff
@@ -0,0 +1,9 @@
+import XCTest
+@testable import Orbit
+
+class SegmentedSwitchTests: SnapshotTestCase {
+
+    func testSegmentedSwitch() {
+        assert(SegmentedSwitchPreviews.snapshot)
```
PavelHolec, line 7: The snapshot reveal some UI issues

- rounded corners have different sizes
- the separator crosses the border

image

Sources/Orbit/Components/SegmentedSwitch.swift
Parabak: For now used `BorderWidth.emphasis` as was suggested for both `borderWidth` and `separatorWidth` constants. No other values are used

Tests/SnapshotTests/Components/SegmentedSwitchTests.swift
Parabak: Extra elevation was added for state with no selection. Fixed

Parabak:
Hey guys, could you please revisit all opened threads, I think all of them are updated.

Tests/SnapshotTests/Components/SegmentedSwitchTests.swift
PavelHolec: Even after the change, the border radius of inner selected view is equal to the outer one - it needs to be smaller to visually match. See 

```
.clipShape(RoundedRectangle(cornerRadius: BorderRadius.default - 1))
```

Parabak: You're right, I haven't check carefully properties in figma. The outer one is 6pts, while inner one is 5. Updated

Sources/Orbit/Components/SegmentedSwitch.swift
```diff
@@ -0,0 +1,214 @@
...
+/// - Note: [Orbit definition](https://orbit.kiwi/components/interaction/segmentedswitch/)
+public struct SegmentedSwitch: View {
+
+    @Binding private var selectedIndex: Int?
+    let label: String
+    let message: Message?
+    let firstOption: String
+    let secondOption: String
+
+    let separatorWidth: CGFloat = BorderWidth.emphasis
```
PavelHolec, line 17: separator will need to have the same size as border, so it does not have to be specified.

Only the optional error outline should use the BorderWidth.emphasis size, otherwise it should be 2pt as specified in design.

Sources/Orbit/Components/SegmentedSwitch.swift
```diff
@@ -0,0 +1,214 @@
...
+                    .layoutPriority(1)
+            }
+        }
+        .animation(.easeOut(duration: 0.2), value: selectedIndex)
+    }
+
+    @ViewBuilder var roundedBackground: some View {
+        RoundedRectangle(cornerRadius: BorderRadius.default)
+            .fill(backgroundColor)
+            .overlay(
```
PavelHolec, line 83: The overlay and `backgroundColor` property should not be necessary - Just the rectangle + fixed `.cloudNormal` color seem to work fine.

Sources/Orbit/Components/SegmentedSwitch.swift
```diff
@@ -0,0 +1,214 @@
...
+    let borderWidth: CGFloat = BorderWidth.emphasis
+
+    public var body: some View {
+        FieldWrapper(
+            label,
+            message: message
+        ) {
+            HStack(spacing: 0) {
+                segment(title: firstOption)
+                    .accessibility(.segmentedSwitchFirstOption)
```
PavelHolec, line 27: Here we can add overlay to the first segment to get rid of `fixedSize`

`.overlay(separator, alignment: .trailing)`

+

`separator...  { Color.cloudNormal.frame(width: borderWidth) }`

Sources/Orbit/Components/SegmentedSwitch.swift
```diff
@@ -0,0 +1,214 @@
...
+        .accessibilityElement(children: .ignore)
+        .accessibility(label: .init(label))
+        .accessibility(value: .init(selection?.description ?? ""))
+        .accessibility(hint: .init(accessibilityHint))
+        .accessibility(addTraits: .isButton)
+        .accessibility(addTraits: selection != nil ? .isSelected : [])
+    }
+
+    @ViewBuilder func segment(title: String) -> some View {
+        Text(title)
```
PavelHolec, line 45: Text will need a center multiline alignment

image

Sources/Orbit/Components/SegmentedSwitch.swift
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    @ViewBuilder var selectionBackground: some View {
+        HStack {
+            if selection == secondOption {
+                Spacer(minLength: 0)
+                    .layoutPriority(1)
+            }
+
+            Color.whiteDarker
```
PavelHolec, line 66: For dark mode, the `.cloudDarkActive` might be a better choice for now

Sources/Orbit/Components/SegmentedSwitch.swift
```diff
@@ -0,0 +1,214 @@
...
+            .previewDisplayName()
+    }
+
+    static var error: some View {
+        segmentedSwitch(message: .error("Error message"))
+            .previewDisplayName()
+    }
+
+    static var interactive: some View {
+        StateWrapper<Int?, _>(initialState: 2) { value in
```
PavelHolec, line 173: `StateWrapper(initialState: Int?(2))` would also work

Sources/Orbit/Components/SegmentedSwitch.swift
```diff
@@ -0,0 +1,214 @@
...
+                SegmentedSwitch(
+                    label: "Gender\nmultiline",
+                    firstOption: "Male with long name",
+                    secondOption: "Female with much longer name",
+                    selectedIndex: value
+                )
+                Text(value.wrappedValue?.description ?? "")
+            }
+        }
+        .previewDisplayName("Live Preview")
```
PavelHolec, line 184: Can be `previewDisplayName()`

Sources/Orbit/Components/SegmentedSwitch.swift
PavelHolec: The hint should only show the message

image

Sources/Orbit/Components/SegmentedSwitch.swift
PavelHolec: The `_ label` is still missing.

Sources/Orbit/Components/SegmentedSwitch.swift
Parabak: Can this be closed as well?
```swift
      segment(title: firstOption)
         .accessibility(.segmentedSwitchFirstOption)
      separator
      segment(title: secondOption)
         .accessibility(.segmentedSwitchSecondOption)
```

Sources/Orbit/Components/SegmentedSwitch.swift
Parabak: Is it up to us? I mean whiteDarker has its own set for light and dark mode. Do we anywhere else do check `UITraitCollection.current.userInterfaceStyle == .dark`?

Sources/Orbit/Components/SegmentedSwitch.swift
Parabak: I've removed `separatorWidth` but left `borderWidth = BorderWidth.emphasis`. If there is an error this is the correct width, and when there is no error, than borderColor is equal to backgroundColor so borderWidth doesn't matter.

Sources/Orbit/Components/SegmentedSwitch.swift
Parabak: I've tried to apply overlay to HStack:
```
.background(selectionBackground)
            .background(roundedBackground)
            .overlay(
                RoundedRectangle(cornerRadius: BorderRadius.default)
                    .strokeBorder(borderColor, lineWidth: borderWidth)
            )
            .fixedSize(horizontal: false, vertical: true)
```
However I've got some weird shadow:
image

Sources/Orbit/Components/SegmentedSwitch.swift
PavelHolec: true, but the non-error separator still needs 2pt width.

Sources/Orbit/Components/SegmentedSwitch.swift
PavelHolec: This change probably needs to be done together with other suggestions.

Sources/Orbit/Components/SegmentedSwitch.swift
PavelHolec: We do, in cases where the color itself is not enough and we need to switch to completely different color for dark mode.

Sources/Orbit/Components/SegmentedSwitch.swift
Parabak: updated

Sources/Orbit/Components/SegmentedSwitch.swift
Parabak: updated

Sources/Orbit/Components/SegmentedSwitch.swift
Parabak: updated

Sources/Orbit/Components/SegmentedSwitch.swift
Parabak: Should be ready as well

Sources/Orbit/Components/SegmentedSwitch.swift
Parabak: updated

Sources/Orbit/Components/SegmentedSwitch.swift
Parabak: in Figma separator is always 2pts in all states. I will use the contstant there 👌

Sources/Orbit/Components/SegmentedSwitch.swift
Parabak: I think I've applied all of them 🤔 except the separator width....

Sources/Orbit/Components/SegmentedSwitch.swift
PavelHolec: The `accessibilityHint` still seems to be construct from values as well. It should only show the (optional) message content.

PavelHolec: Is the change pushed? The `accessibilityHint` still seems to be construct from values as well. It should only show the (optional) message content.

Sources/Orbit/Components/SegmentedSwitch.swift
Parabak: found issue, will be updated

Sources/Orbit/Components/SegmentedSwitch.swift
PavelHolec: I realized it is probably trickier. This would not be accessible anyway, as we do not expose both buttons with their values. Lets keep it as-is for now, please create a follow-up task.

Parabak: Hint is applied on top of `FieldWrapper`
```
   .accessibility(hint: .init(accessibilityHint))
   
   ...
   
    private var accessibilityHint: String {
        "\(message?.description.appending(":") ?? "") \(firstOption)\\\(secondOption)"
    }
```
image
Screenshot 2023-01-06 at 13 03 38

Is it wrong? I can create a separate task but what would be the ideal value?

PavelHolec: The hint should only specify the "message", as in other controls, not values. But a follow-up task is necessary anyway, as this component needs to mimic the accessibility of segmented control.

Parabak: Issue is created here https://github.com/kiwicom/orbit-swiftui/issues/417

Parabak:
All  changes are pushed

PavelHolec (approved):

PR #409, opened 2022-12-28
Unify `InputField` sizing and fix `SecureTextField` size

PavelHolec:
image

#### Dynamic type

image

image


Sources/Orbit/Support/Components/SecureTextField.swift
```diff
@@ -9,6 +9,23 @@ struct SecureTextFieldStyle {
     let state: InputState
 }
 
+private class InsettedTextField: UITextField {
```
sjavora, line 12: `InsetTextField` - probably doesn't matter, but `set` is not a regular English verb... it's the same in all tenses.

Sources/Orbit/Support/Previews/PreviewFormatting.swift
```diff
@@ -0,0 +1,10 @@
+import Foundation
+
+extension CGFloat {
+
+    var formatted: String {
+        -0.001...0.001 ~= self.remainder(dividingBy: 1)
```
sjavora, line 6: Is that really more readable than `(-0.001...0.001).contains(self.remainder(dividingBy: 1))`? 🤔 

Or even `abs(self.remainder(dividingBy: 1)) <= 0.001`?

sjavora:
`.measure` is nice 👌 

I thought about suggesting something like `.measured()` instead, but not sure now what the conventions within SwiftUI are.

sjavora:
Any idea why input fields snapshots say 43 instead of 44 shown in the PR description?

Sources/Orbit/Support/Previews/PreviewFormatting.swift
PavelHolec: Updated using the `abs` version

PavelHolec:
> `.measure` is nice 👌
> 
> I thought about suggesting something like `.measured()` instead, but not sure now what the conventions within SwiftUI are.

Sounds better, renamed. If a horizontal measurement will be needed in future, it might become `measured(axis:)` with default value set to `.vertical`

PavelHolec:
> Any idea why input fields snapshots say 43 instead of 44 shown in the PR description?

It is due to the default font having different size. A follow-up #405 will fix this by having fixed (designed) line height independently on used font.

Sources/Orbit/Support/Components/SecureTextField.swift
hrach: https://en.m.wiktionary.org/wiki/insetted
is it or do I get it wrong?

sjavora: There seems to be precious few references to that form... with quotes, I only get 169k hits on everything indexed by Google (and I still get "did you mean inserted?") as opposed to 104m for "inset". Granted, those would also contain the noun.

Meanwhile, these sites don't show "insetted":
- https://www.thefreedictionary.com/insetted (redirects to "inset")
- https://www.oxfordlearnersdictionaries.com/definition/english/inset_2

even https://en.wiktionary.org/wiki/inset shows it as "inset or insetted".

Even discounting all that though, playing with the prefix IMO shows that it should be "inset"... there is no "resetted" for example - just "reset". wiktionary of course includes "setted" for some reason, but doesn't even have a citation for that one. I'm beginning to suspect it as a reliable source now...

hrach: I have the feeling that I've seen it before.
https://github.com/search?q=insetted&type=code
this is imo more related than google search.
but no issue with inset.

sjavora: Too bad we can't distinguish nouns here either, because just for "inset", I get 43 million results as opposed to less than 10k for "insetted".

> this is imo more related than google search.

Yeah, but should programmers be the authority on natural language? 😅 I'm willing to bet a lot of them are not native speakers - that's why I put the dictionary links in my reply.

sjavora (approved):

PR #418, opened 2023-01-06
Add inline `Alert`

### LINKED ISSUE #75: Support `.inline` variant for `Alert`
https://www.figma.com/file/YMfJkmUMV8xiY7fgeW6QAr/%5BOrbit%5D-Native-Components?node-id=7037%3A4113

sjavora:
Some duplication is creeping in, but ultimately I stuck with only having a single `Alert` component instead of creating a separate `InlineAlert`.

Screenshot 2023-01-06 at 22 21 23
Screenshot 2023-01-06 at 22 21 33
Screenshot 2023-01-06 at 22 21 54

Closes #75.


Sources/Orbit/Components/Alert.swift
```diff
@@ -30,33 +30,44 @@ public struct Alert<Content: View>: View {
...
+
+    @ViewBuilder var content: some View {
+        switch style {
+            case .default:  defaultContent
+            case .inline:   inlineContent
+        }
+    }
+
+    @ViewBuilder var defaultContent: some View {
+        HStack(alignment: .firstTextBaseline, spacing: .xSmall) {
```
PavelHolec, line 59: For both `defaultContent` and `inlineContent`, the `.top` alignment is more precise and preferred for constructing Orbit components (so it matches design heights 1:1). 

The `.firstTextBaseline` alignment for Icon+Text is approximate and is meant for use where Icon and Text sizes (Icon size and text line height to be exact) do not match.

PavelHolec:
Looks great! 🎊

Sources/Orbit/Components/Alert.swift
sjavora: `.top` really doesn't look good for single-line content though.

With `.firstTextBaseline`:
Screenshot 2023-01-07 at 9 17 27

With `.top`
Screenshot 2023-01-07 at 9 17 15

PavelHolec: True. Designs use `centered` as they use single-line example, which is not useful. But they visually seem to match the baseline alignment 👍 

The leading icon + text should still be stacked together using `top` alignment, with the inline button being aligned to `firstbaseline` in outer HStack.

PavelHolec:
The update makes intrinsic size not respected anymore (`idealSize` environment key)

image

PavelHolec:
Inline variant can now be added to `Storybook`

sjavora:
> Inline variant can now be added to `Storybook`

They are - see changes to `StorybookAlert.swift` and `StorybookDetail.swift`.

Sources/Orbit/Components/Alert.swift
sjavora: Ok, two stacks it is.

Sources/Orbit/Components/Alert.swift
```diff
@@ -242,13 +317,19 @@ struct AlertPreviews: PreviewProvider {
...
+            Alert(
+                "Alert with very very very very very very very very very very very long and <u>multiline</u> title",
+                description: description,
+                icon: .grid,
+                buttons: primaryAndSecondaryConfiguration
+            ) {
+                contentPlaceholder
+            }
+
+            Alert("Alert with very very very", icon: .grid, button: "Primary", status: .warning, isSuppressed: false)
```
PavelHolec, line 330: This one is probably meant to include the `.idealSize(horizontal: true)` as these are now the same in snapshots?

sjavora: Oh, this is an oversight from adding the one below - this one is the "default" inline alert and was previously named that - `Inline alert`. I'll revert the name.

PR #421, opened 2023-01-11
Fix ChoiceTile content padding

PavelHolec:
image

image


sjavora:
Is the badge shift intended/wanted?

Screenshot 2023-01-11 at 16 17 29

PavelHolec:
> Is the badge shift intended/wanted?
> 
> Screenshot 2023-01-11 at 16 17 29

Should be fixed now. Updated the API to be more consistent with ListChoice, removing the (non-Orbit) badge and replaced by any custom header content.

Sources/Orbit/Components/ChoiceTile.swift
```diff
@@ -186,12 +199,23 @@ public struct ChoiceTile<Content: View>: View {
         }
     }
     
-    var messagePadding: CGFloat {
-        indicatorSize > 0
-            ? (indicatorSize + padding + .xSmall)
+    var indicatorTrailingPadding: CGFloat {
+//        indicatorSize > 0 && alignment == .default && headerContent is EmptyView
```
sjavora, line 203: Commented code.

sjavora:
Also conflicts.

Sources/Orbit/Components/ChoiceTile.swift
PavelHolec: Thx, updated

sjavora (approved):

PR #443, opened 2023-01-19
Update inline Alert layout to match design

PavelHolec:
The variant with no buttons was disambiguous between normal and inline layout (the new `init` seems to make the inline variant preferred in case of no buttons and a title-only). Updated to match designs and some edge cases (only title, only button)

image

#### Updated 

image


Sources/Orbit/Components/Alert.swift
```diff
@@ -79,32 +78,37 @@ public struct Alert<Content: View>: View {
...
+            }
+
+            if idealSize.horizontal != true {
+                Spacer(minLength: 0)
             }
 
             switch buttons {
                 case .primary(let button),
-                     .primaryAndSecondary(let button, _):
+                        .primaryAndSecondary(let button, _):
```
sjavora, line 102: Can we keep these aligned?

Sources/Orbit/Components/Alert.swift
```diff
@@ -191,6 +195,10 @@ public struct Alert<Content: View>: View {
             ? .secondary
             : .status(status, subtle: true)
     }
+
+    var isHeaderEmpty: Bool {
```
sjavora, line 199: I think I got rid of this on purpose - does it actually do anything if all the parts resolve to `EmptyView` anyway?

Sources/Orbit/Components/Alert.swift
PavelHolec: Yes, it is probably Xcode indenting

Sources/Orbit/Components/Alert.swift
PavelHolec: It seems to be required, because nested stack views do not seem to have the recursive logic for evaluating whether the whole content tree consists purely of `EmptyView`s up to the root stack. This means, there would be empty "stacks" that take part in parent stack spacing.

PavelHolec:
@sjavora refactored to extract properties from `default` and `inline` content.

sjavora (approved):

PR #449, opened 2023-01-21
Remove default values for component actions

### LINKED ISSUE #437: Remove default values for component actions
Consider removing default empty `{}` action handlers for all components and make it mandatory. This may prevent the disambiguity when combined with multiple viewbuilder closures.

sjavora:
I left default values for text link actions, all others should be removed.

Closes #437.


PavelHolec (approved):
Nice 👍 

The actions show some `missing value for key 'id' of option 'Destination'` error - probably from snapshot test build?

Issue #456, opened 2023-01-25
Improve Slider accessibility 


Parabak:
* Adjust slider via Accisibility as SwiftUI.Slider does
* Check RTL Locales

PR #457, opened 2023-01-25
`Slider` component

Parabak:
Slider component Implementation

Created a [standlone issue](https://github.com/kiwicom/orbit-swiftui/issues/456) for accessibility improvements


sjavora:
@Parabak please rebase - you'll see why CI can't build.

Sources/Orbit/Components/Slider.swift
```diff
@@ -0,0 +1,421 @@
...
+                            onValueChange(bounds.upperBound)
+                            return
+                        }
+
+                        guard newHorisontalPosition > 0 else {
+                            onValueChange(bounds.lowerBound)
+                            return
+                        }
+
+                        if step != nil {
```
sjavora, line 148: `step` isn't used in this `if` - it seems disconnected - why is `step` the deciding factor here? Shouldn't we use `buckets.isEmpty == false` or something like that?

Sources/Orbit/Components/Slider.swift
```diff
@@ -0,0 +1,421 @@
...
+    private var lineCenter: CGFloat {
+        lineHeight / 2
+    }
+
+    private var thumbOverlapping: CGFloat {
+        (SliderThumb.circleDiameter / 2 - lineCenter) * sizeCategory.controlRatio
+    }
+
+    private var accessibilityHint: String {
+        startLabel + "—" + endLabel
```
sjavora, line 185: Don't we need some spaces around the dash?

Also I'd probably use string interpolation.

Sources/Orbit/Components/Slider.swift
```diff
@@ -0,0 +1,421 @@
...
+    }
+
+    private func updateSelectedRangeIfNeeded() {
+        if let upperBound = secondValue, case let .range(selectedRange) = value {
+            selectedRange?.wrappedValue = min(firstValue, upperBound)...max(firstValue, upperBound)
+        }
+    }
+
+    private func currentValue(location: CGFloat, sliderWidth: CGFloat) -> Value {
+        return bounds.lowerBound + bounds.upperBound * Value(location / sliderWidth)
```
sjavora, line 195: Unnecessary `return`

Sources/Orbit/Components/Slider.swift
```diff
@@ -0,0 +1,421 @@
...
+    }
+
+    private func position(by value: Value, sliderWidth: CGFloat) -> CGFloat {
+        let position = CGFloat(value / bounds.upperBound) * sliderWidth
+        let normalisied = min(sliderWidth, max(0, position))
+        return normalisied
```
sjavora, line 201: You can `return` on the previous line.

Also, normali**z**ed.

Sources/Orbit/Components/Slider.swift
```diff
@@ -0,0 +1,421 @@
...
+
+    public init(
+        selectedRange: Binding<ClosedRange<Value>>,
+        valueLabel: String,
+        startLabel: String,
+        endLabel: String,
+        bounds: ClosedRange<Value>,
+        step: Value.Stride? = nil
+    ) {
+        assert(step != 0, "step can'be 0")
```
sjavora, line 231: can't be

Sources/Orbit/Components/Slider.swift
```diff
@@ -0,0 +1,421 @@
...
+    var circleDiameter: CGFloat {
+        Self.circleDiameter * sizeCategory.controlRatio
+    }
+
+    var dotDiameter: CGFloat {
+        Self.dotDiameter * sizeCategory.ratio
+    }
+}
+
+extension ClosedRange where Bound: BinaryFloatingPoint {
```
sjavora, line 277: `extension ClosedRange<BinaryFloatingPoint>`

Sources/Orbit/Components/Slider.swift
```diff
@@ -0,0 +1,421 @@
...
+    }
+}
+
+extension Range where Bound: BinaryFloatingPoint {
+    var middle: Bound {
+        lowerBound + (upperBound - lowerBound) / 2
+    }
+}
+
+extension Slider where Value == Double {
```
sjavora, line 291: `extension Slider<Double>`

Sources/Orbit/Components/Slider.swift
```diff
@@ -0,0 +1,421 @@
...
+    init(value: Binding<Value>, bounds: ClosedRange<Value>, step: Value.Stride? = nil) {
+        self.init(
+            value: value,
+            valueLabel: value.wrappedValue.formatted,
+            endLabel: bounds.upperBound.formatted,
```
sjavora, line 296: I wonder what this `formatted` does in various locales 🤔

Sources/Orbit/Components/Slider.swift
```diff
@@ -0,0 +1,421 @@
...
+            startLabel: bounds.lowerBound.formatted,
+            endLabel: bounds.upperBound.formatted,
+            bounds: bounds, step: step)
+    }
+}
+
+// MARK: - Previews
+struct SliderPreviews: PreviewProvider {
+
+    // TODO: snapshots
```
sjavora, line 316: TODO

Sources/Orbit/Components/Slider.swift
```diff
@@ -0,0 +1,421 @@
...
+            .foregroundColor(Color.blueNormal)
+            .frame(width: dotDiameter, height: dotDiameter)
+    }
+
+    var circleDiameter: CGFloat {
+        Self.circleDiameter * sizeCategory.controlRatio
+    }
+
+    var dotDiameter: CGFloat {
+        Self.dotDiameter * sizeCategory.ratio
```
PavelHolec, line 273: Probably the `controlRatio` should be used to resize the same way

Sources/Orbit/Components/Slider.swift
PavelHolec: We should avoid formatting it at Orbit level. The Orbit should take exactly what the call size wants to display, 1:1

Sources/Orbit/Components/Slider.swift
```diff
@@ -0,0 +1,421 @@
...
+
+    }
+
+    @ViewBuilder func thumb(
+        horisontalPosition: CGFloat,
+        sliderWidth: CGFloat,
+        isDragging: Binding<Bool>,
+        onValueChange: @escaping (Value) -> Void
+    ) -> some View {
+        SliderThumb()
```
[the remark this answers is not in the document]
hrach: Thumb is the common name in the wild how to call this.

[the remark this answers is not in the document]
sjavora: Even by Apple ([UIKit](https://developer.apple.com/documentation/uikit/uislider), [SwiftUI](https://developer.apple.com/documentation/swiftui/slider)):

14f9f16e-55c0-474a-ae62-f42ebf2cab33

Sources/Orbit/Components/Slider.swift
Parabak: It's the `internal init` visible and used only in  Previews, so it won't affect users

Sources/Orbit/Components/Slider.swift
```diff
@@ -0,0 +1,421 @@
...
+    let value: SelectedValue
+    let valueLabel: String
+    let startLabel: String
+    let endLabel: String
+    let bounds: ClosedRange<Value>
+    let step: Value.Stride?
+
+    @State private var isFirstThumbDrag = false
+    @State private var isSecondThumbDrag = false
+    @State private var firstValue: Value {
```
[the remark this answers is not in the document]
Parabak: I've just read some articles that Apple uses this way to emphasize that @State has a private nature...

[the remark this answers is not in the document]
sjavora: Maybe it's time we establish a code style for Orbit then... I'm going applying Kiwi rules here.

Sources/Orbit/Components/Slider.swift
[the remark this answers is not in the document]
PavelHolec: We usually try to mimic the Orbit and React naming to remove inconsistencies though 🤷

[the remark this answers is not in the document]
sjavora: Then maybe someone should tell Orbit not to rock the boat 😄

Sources/Orbit/Components/Slider.swift
PavelHolec: But those previews can be used as examples by devs. We should not use internal convenience inits in previews, but the real API, so that we can even test that the API is easy to use.

Sources/Orbit/Components/Slider.swift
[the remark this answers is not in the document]
PavelHolec: Yep, they use it to teach and emphasize it, which is a good thing for newcomers. In real usage, it becomes a boilerplate.

Sources/Orbit/Components/Slider.swift
```diff
@@ -0,0 +1,421 @@
+import SwiftUI
+
+public typealias SliderValue = BinaryFloatingPoint & Strideable & CustomStringConvertible
+
+/// Enables selection from a range of values.
+///
+/// - Note: [Orbit definition](https://orbit.kiwi/components/interaction/slider/)
+public struct Slider<Value>: View where Value: SliderValue, Value.Stride: BinaryFloatingPoint {
```
Parabak, line 8: Slider design was update in Figma, so the implementation is outdated

PetroShylaiev:
Closing as was merged in a separate PR:
https://github.com/kiwicom/orbit-swiftui/pull/474

Sources/Orbit/Components/Slider.swift
```diff
@@ -0,0 +1,441 @@
...
+    private func currentValue(location: CGFloat, width: CGFloat) -> Value {
+        bounds.lowerBound + Value(location / width) * bounds.distance
+    }
+
+    private func position(by value: Value, width: CGFloat) -> CGFloat {
+        let position = CGFloat((value - bounds.lowerBound) / bounds.distance) * width
+        return min(width, max(0, position))
+    }
+
+    public init(
```
PavelHolec, line 213: Lets extract inits to `public extension` and put comments to them so we get nice docs. Should be done consistently with other component inits.

Parabak: Updated.

Sources/Orbit/Components/Slider.swift
```diff
@@ -0,0 +1,441 @@
...
+        value: Binding<Value>,
+        valueLabel: String,
+        startLabel: String,
+        endLabel: String,
+        bounds: ClosedRange<Value>,
+        step: Value.Stride? = nil
```
PavelHolec, line 219: If possible, the step value could be used for rounding the value so that call site does not have to clean it:

image

Sources/Orbit/Components/Slider.swift
```diff
@@ -0,0 +1,441 @@
...
+    }
+
+    private func position(by value: Value, width: CGFloat) -> CGFloat {
+        let position = CGFloat((value - bounds.lowerBound) / bounds.distance) * width
+        return min(width, max(0, position))
+    }
+
+    public init(
+        value: Binding<Value>,
+        valueLabel: String,
```
PavelHolec, line 215: As in with other components, label values should be empty by default, so the call site does not have to fill every parameter

Sources/Orbit/Components/Slider.swift
```diff
@@ -0,0 +1,441 @@
...
+        self.endLabel = endLabel
+
+        self.bounds = bounds
+
+        self._firstValue = State(initialValue: value.wrappedValue)
+        self.buckets = bounds.chunked(into: step ?? Value.Stride(bounds.upperBound - bounds.lowerBound))
+    }
+
+    public init(
+        selectedRange: Binding<ClosedRange<Value>>,
```
PavelHolec, line 235: Suggestion - to match the other init, this can be named `values` or `valueRange`

Sources/Orbit/Components/Slider.swift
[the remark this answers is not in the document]
PavelHolec: Does it mean the Slider does not support the Int range? 🤔 

Now it says:
```
Generic struct 'Slider' requires that 'Int.Stride' (aka 'Int') conform to 'BinaryFloatingPoint'
```

Sources/Orbit/Components/Slider.swift
```diff
@@ -0,0 +1,441 @@
...
+    }
+
+    @State private var dragStartOffset = 0.0
+
+    private let buckets: [Range<Value>]
+
+    public var body: some View {
+        VStack(alignment: .leading, spacing: 0) {
+            Text(valueLabel)
+                .foregroundColor(.blueNormal)
```
PavelHolec, line 49: The foregroundColor now needs to be replaced with `textColor`

Sources/Orbit/Components/Slider.swift
Parabak: I will add it but I believe this is a bit different case. A slider without any UI representation of its value is strange.

Sources/Orbit/Components/Slider.swift
```diff
@@ -0,0 +1,421 @@
...
+
+/// Enables selection from a range of values.
+///
+/// - Note: [Orbit definition](https://orbit.kiwi/components/interaction/slider/)
+public struct Slider<Value>: View where Value: SliderValue, Value.Stride: BinaryFloatingPoint {
+
+    enum SelectedValue {
+        case single(Binding<Value>?)
+        case range(Binding<ClosedRange<Value>>?)
+    }
```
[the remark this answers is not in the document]
PavelHolec: Wouldn't that be unnecessary complicated for a call site though? Most of the time, a call site might use just one or the other slider and would want to map the value directly to some numerical state, without having to map it via custom binding.

Sources/Orbit/Components/Slider.swift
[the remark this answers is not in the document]
PavelHolec: On the other hand, we already started to have components that have builder init properties. Using those, any internal properties (environment, state, or others) show in the autocomplete (when working on the library). Probably not an issue, but we should unify it in a follow-up and keep it as-is here.

Sources/Orbit/Components/Slider.swift
```diff
@@ -0,0 +1,441 @@
...
+                case .single(let currentValue):
+                    currentValue?.wrappedValue = firstValue
+                case .range:
+                    updateSelectedRangeIfNeeded()
+            }
+        }
+    }
+
+    @State private var secondValue: Value? {
+        didSet {
```
PavelHolec, line 37: Do these `didSet` work? AFAIK this is not supposed to be working 🤔 

https://forums.swift.org/t/state-vars-didset-fix-or-prohibition/53161

https://developer.apple.com/forums/thread/123357

Sources/Orbit/Components/Slider.swift
```diff
@@ -0,0 +1,441 @@
...
+
+            SingleAxisGeometryReader(axis: .horizontal, alignment: .bottom) { width in
+                track
+                    .overlay(
+                        controls(sliderWidth: width)
+                    )
+                    .padding(.vertical, thumbVerticalOverlapping)
+            }
+
+            HStack {
```
PavelHolec, line 61: The stack is missing both spacing and alignment. The Spacer also misses the `minLength` property. This would result in a layout that is out of our control.

Sources/Orbit/Components/Slider.swift
```diff
@@ -0,0 +1,441 @@
...
+            }
+            .padding(.top, .xxSmall)
+            .accessibilityElement(children: .ignore)
+            .accessibility(label: .init(valueLabel))
+            .accessibility(hint: .init(accessibilityHint))
+        }
+    }
+
+    @ViewBuilder var track: some View {
+        RoundedRectangle(cornerRadius: 100)
```
PavelHolec, line 77: 100? 🤔

Sources/Orbit/Components/Slider.swift
```diff
@@ -0,0 +1,441 @@
...
+            .fill(Color.cloudDark)
```
PavelHolec, line 78: Now we can use the syntax `.fill(.cloudDark)`

Sources/Orbit/Components/Slider.swift
```diff
@@ -0,0 +1,441 @@
...
+        self.bounds = bounds
+
+        self._firstValue = State(initialValue: selectedRange.wrappedValue.lowerBound)
+        self._secondValue = State(initialValue: selectedRange.wrappedValue.upperBound)
+
+        self.buckets = bounds.chunked(into: step ?? Value.Stride(bounds.upperBound))
+    }
+}
+
+struct SliderThumb: View {
```
PavelHolec, line 257: Should be separated into Support/Components

Sources/Orbit/Components/Slider.swift
```diff
@@ -0,0 +1,441 @@
...
+            updateSelectedRangeIfNeeded()
+        }
+    }
+
+    @State private var dragStartOffset = 0.0
+
+    private let buckets: [Range<Value>]
+
+    public var body: some View {
+        VStack(alignment: .leading, spacing: 0) {
```
PavelHolec, line 47: The leading side seems to have wrong rounding of selected area

image

Sources/Orbit/Components/Slider.swift
```diff
@@ -0,0 +1,441 @@
...
+///
+/// - Note: [Orbit definition](https://orbit.kiwi/components/interaction/slider/)
+public struct Slider<Value>: View where Value: SliderValue, Value.Stride: BinaryFloatingPoint {
+
+    enum SelectedValue {
+        case single(Binding<Value>?)
+        case range(Binding<ClosedRange<Value>>?)
+    }
+
+    @Environment(\.sizeCategory) var sizeCategory
```
PavelHolec, line 15: The sizeCategory has some glitches. Not only when the slider causes the surrounding context to be changed as a result, but even in the initialize state

https://user-images.githubusercontent.com/35844477/235114460-0c0212a5-2dee-48a1-b643-d3d430bc8e86.mov

Sources/Orbit/Components/Slider.swift
Parabak: `Step` is an optional parameter, there is no default value 🤷 

Looks like I can't find a way how to fix this. The SwiftUI slider has the same issue and this issue is enevitable with working with doubles. So I would leave it up to client. Working with BinaryFloatingPoint doesn't allow much rounding either. We can round to  Int  but not to some fraciton

I've tried to use a Decimal type instead of BinaryFloatingPoint. This resolved issue with buckets (the print output):
```swift
[
Range(0..<10.3), Range(10.3..<20.6), Range(20.6..<30.9), Range(30.9..<41.2), Range(41.2..<51.5), 
Range(51.5..<61.8), Range(61.8..<72.1), Range(72.1..<82.4), Range(82.4..<92.7), Range(92.7..<100)
]
```
However converting from Decimal to Int/Double/Float will be more tricker. I actually don't even know the way except of using obj-C NSDecimalNumber for that 🤷 

So decimal gives us more precision but will introduce complications with actual component usage

Sources/Orbit/Components/Slider.swift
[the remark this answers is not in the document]
Parabak: What Pavel wrote is what I thought when I've implemented it.

Sources/Orbit/Components/Slider.swift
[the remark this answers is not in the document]
Parabak: @PavelHolec so you're agreed to keep it private for Orbit?

Sources/Orbit/Components/Slider.swift
```diff
@@ -0,0 +1,421 @@
...
+    @ViewBuilder func controls(width: CGFloat) -> some View {
+
+        let firstThumbPosition = position(by: firstValue, sliderWidth: width)
+
+        switch secondValue {
+            case nil:
+                firstThumb(atPosition: firstThumbPosition, sliderWidth: width)
+                    .background(strokedPath(from: 0, to: firstThumbPosition))
+
+            case .some(let secondValue):
```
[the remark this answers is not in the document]
Parabak: can be closed as outdated

Sources/Orbit/Components/Slider.swift
```diff
@@ -0,0 +1,421 @@
...
+
+    @ViewBuilder func secondThumb(atPosition horisontalPosition: CGFloat, sliderWidth: CGFloat) -> some View {
+        thumb(
+            horisontalPosition: horisontalPosition,
+            sliderWidth: sliderWidth,
+            isDragging: $isSecondThumbDrag
+        ) {
+            self.secondValue = $0
+        }
+        .zIndex(isSecondThumbDrag ? 1 : 0)
```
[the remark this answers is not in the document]
Parabak: Still works 😄

Sources/Orbit/Components/Slider.swift
Parabak: Is it really the same? In case of you case compiler complains about `Binary operator '-' cannot be applied to two 'Bound' operands`

Sources/Orbit/Components/Slider.swift
Parabak: From your first link:
>  So it is somewhat expected in that perspective for the didSet to only trigger on direct mutations of the property itself (and not through it's projection).

Also here https://forums.swift.org/t/using-error-protocol-in-swiftui-binding-does-any-or-some-fix-this/62740/4:
> The reason is that this didSet will trigger only via the computed error.set accessor and the Binding generated by State does not pass the value through that. It does only work if the view sets the property directly (e.g. error = someError), not through a binding.

This is our case.

Sources/Orbit/Components/Slider.swift
Parabak: I've copied from Figma. I didn't want to introduce anything from my side even though 100 looks wrong
image

PavelHolec: It does look wrong. A workaround or hack in figma does not necessarily translate to SwiftUI - both layouts have their own gotchas. Lets use a proper radius.

Sources/Orbit/Components/Slider.swift
PavelHolec: I still cannot see the change in gitlab 🤔

Sources/Orbit/Components/Slider.swift
Parabak: Nice catch, missed that. Applied cornerRadius on progress as well.

Sources/Orbit/Components/Slider.swift
Parabak: GitHub, right? It's not yet pushed😅 I will ping you back once all threads will be updated / commented.

Sorry, GitHub doesn't have this fancy GitLab's Start Review feature, so I can notify you only when all threads are updated.

Parabak:
@PavelHolec, @sjavora all threads should be updated / commented. Please have a look

Parabak:
Weird thing, CI complains about wrong type `Int` but there is no issue when I build locally... Do you know anything about it ?What Xcode version do we use on CI?

sjavora:
Probably 14.2, but this doesn't look like that type of issue. Can you still build locally if you clone a fresh repo?

PavelHolec (approved):

PR #463, opened 2023-02-01
`Stepper` component

### LINKED ISSUE #78: Create `Stepper` component
Including `StepperButtonsStyle`.

https://www.figma.com/file/YMfJkmUMV8xiY7fgeW6QAr/%5BOrbit%5D-Native-Components?node-id=8086%3A3580

PetroShylaiev:
Closes #78

### DIFF
--- a/Sources/Orbit/Components/Stepper.swift
+++ b/Sources/Orbit/Components/Stepper.swift
@@ -1,0 +1,239 @@
+import SwiftUI
+
+/// Enables incremental changes of a counter without a direct input.
+///
+/// - Note: [Orbit definition](https://orbit.kiwi/components/interaction/stepper/)
+public struct Stepper: View {
+    
+    @Binding var value: Int
+    
+    let minValue: Int
+    let maxValue: Int
+    let style: Style
+    
+    @Environment(\.isEnabled) var isEnabled
+    
+    public var body: some View {
+        HStack(alignment: .center, spacing: 0) {
+            decrementButton
+            
+            valueText
+                .frame(minWidth: 40)
+            
+            incrementButton
+        }
+    }
+    
+    @ViewBuilder var valueText: some View {
+        Text("\(value)")
+            .accessibility(.stepperValue)
+    }
+    
+    @ViewBuilder var decrementButton: some View {
+        StepperButton(
+            isIncrement: false,
+            isEnabled: isEnabled ? value > minValue : false,
+            style: style,
+            action: {
+                value -= 1
+            }
+        )
+        .accessibility(.stepperDecrement)
+    }
+    
+    @ViewBuilder var incrementButton: some View {
+        StepperButton(
+            isIncrement: true,
+            isEnabled: isEnabled ? value < maxValue : false,
+            style: style,
+            action: {
+                value += 1
+            }
+        )
+        .accessibility(.stepperIncrement)
+    }
+    
+    /// Creates Orbit Stepper component.
+    public init(
+        value: Binding<Int>,
+        minValue: Int,
+        maxValue: Int,
+        style: Style
+    ) {
+        self._value = value
+        self.minValue = minValue
+        self.maxValue = maxValue
+        self.style = style
+    }
+}
+
+// MARK: - Styles
+extension Stepper {
+    
+    public enum Style {
+        case primary
+        case secondary
+        
+        public var foregroundColor: Color {
+            switch self {
+            case .primary:                  return .white.opacity(0.5)
+            case .secondary:                return .inkDark.opacity(0.5)
+            }
+        }
+        
+        public var foregroundActiveColor: Color {
+            switch self {
+            case .primary:                  return .white
+            case .secondary:                return .inkDark
+            }
+        }
+        
+        public var backgroundColor: Color {
+            switch self {
+            case .primary:                  return .blueNormal.opacity(0.5)
+            case .secondary:                return .cloudNormal.opacity(0.5)
+            }
+        }
+        
+        public var backgroundActiveColor: Color {
+            switch self {
+            case .primary:                  return .blueNormal
+            case .secondary:                return .cloudNormal
+            }
+        }
+        
+        public var borderSelectedColor: Color {
+            switch self {
+            case .primary:                  return .blueLightActive
+            case .secondary:                return .cloudNormalActive
+            }
+        }
+    }
+}
+
+// MARK: - Identifiers
+public extension AccessibilityID {
+    
+    static let stepperIncrement         = Self(rawValue: "orbit.stepper.increment")
+    static let stepperDecrement         = Self(rawValue: "orbit.stepper.decrement")
+    static let stepperValue             = Self(rawValue: "orbit.stepper.value")
+}
+
+// MARK: - Previews
+struct StepperPreviews: PreviewProvider {
+    
+    static var previews: some View {
+        PreviewWrapper {
+            standalone
+            states
+            statesLargeFont
+        }
+        .padding(.medium)
+        .previewLayout(.sizeThatFits)
+    }
+    
+    static var zeroToTen: some View {
+        StateWrapper(3) { binding in
+            Stepper(
+                value: binding,
+                minValue: -4,
+                maxValue: 10,
+                style: .primary
+            )
+        }
+    }
+    
+    static var minusFiveToTwenty: some View {
+        StateWrapper(-3) { binding in
+            Stepper(
+                value: binding,
+                minValue: -5,
+                maxValue: 20,
+                style: .primary
+            )
+        }
+    }
+    
+    static var twoToFifteen: some View {
+        StateWrapper(2) { binding in
+            Stepper(
+                value: binding,
+                minValue: 2,
+                maxValue: 15,
+                style: .primary
+            )
+        }
+    }
+    
+    static var zeroToThree: some View {
+        StateWrapper(2) { binding in
+            Stepper(
+                value: binding,
+                minValue: 0,
+                maxValue: 3,
+                style: .primary
+            )
+        }
+    }
+    
+    static var disabledZeroToThree: some View {
+        StateWrapper(2) { binding in
+            Stepper(
+                value: binding,
+                minValue: 0,
+                maxValue: 3,
+                style: .primary
+            )
+            .disabled(true)
+        }
+    }
+    
+    static var secondaryThreeToTen: some View {
+        StateWrapper(5) { binding in
+            Stepper(
+                value: binding,
+                minValue: 3,
+                maxValue: 10,
+                style: .secondary
+            )
+        }
+    }
+    
+    static var standalone: some View {
+        zeroToTen
+            .previewDisplayName()
+    }
+    
+    static var states: some View {
+        VStack(spacing: .large) {
+            minusFiveToTwenty
+            twoToFifteen
+            zeroToThree
+            disabledZeroToThree
+            secondaryThreeToTen
+        }
+        .previewDisplayName()
+    }
+    
+    static var statesLargeFont: some View {
+        VStack(spacing: .large) {
+            minusFiveToTwenty
+                .environment(\.sizeCategory, .extraExtraExtraLarge)
+            disabledZeroToThree
+                .environment(\.sizeCategory, .extraExtraExtraLarge)
+            secondaryThreeToTen
+                .environment(\.sizeCategory, .extraExtraExtraLarge)
+        }
+        .previewDisplayName()
+    }
+    
+    static var snapshot: some View {
+        states
+            .padding(.medium)
+    }
+    
+    static var snapshotLargeFont: some View {
+        statesLargeFont
+            .padding(.medium)
+    }
+}
--- a/Sources/Orbit/Orbit.docc/Uncategorized.md
+++ b/Sources/Orbit/Orbit.docc/Uncategorized.md
@@ -50,6 +50,7 @@ Supporting types and components under development.
 - ``PasswordStrengthIndicator``
 - ``Progress``
 - ``Status``
+- ``StepperButton``
 
 ### Deprecated
 
--- a/Sources/Orbit/Support/Components/StepperButton.swift
+++ b/Sources/Orbit/Support/Components/StepperButton.swift
@@ -1,0 +1,49 @@
+import SwiftUI
+
+/// An icon-based stepper button for suitable for actions inside components like `Stepper`
+public struct StepperButton: View {
+    
+    let isIncrement: Bool
+    let isEnabled: Bool
+    let style: Stepper.Style
+    let action: () -> Void
+    
+    public var body: some View {
+        SwiftUI.Button(action: action) {
+            icon
+        }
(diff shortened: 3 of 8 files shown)


PavelHolec (changes requested):
Missing:

- Storybook updates
- snapshot tests
- component status (released)

PetroShylaiev:
> * Storybook updates
Added
> * snapshot tests
Added
> * accessibility and identifiers
Added
> * component status (released)
Changed

Sources/Orbit/Components/Stepper.swift
```diff
@@ -0,0 +1,262 @@
...
+            }
+        )
+    }
+
+    /// Creates Orbit Stepper component.
+    public init(
+        value: Binding<Int>,
+        minValue: Int,
+        maxValue: Int,
+        isEnabled: Bool = true,
```
sjavora, line 81: We're now using the native `@Environment(\.isEnabled)` for this.

Sources/Orbit/Components/Stepper.swift
```diff
@@ -0,0 +1,262 @@
...
+        self.style = style
+    }
+}
+
+// MARK: - Subtypes
+extension Stepper {
+
+    struct StepperButton: View {
+
+        let isIncrement: Bool
```
sjavora, line 97: This only selects which icon is used - I'd probably replace this property with the icon, since this is only accessible in this file, the danger of some random icon being assigned is very low.

Sources/Orbit/Components/Stepper.swift
```diff
@@ -0,0 +1,262 @@
...
+                isIncreased = false
+
+                withAnimation {
+                    value -= 1
+                }
+            }
+        )
+    }
+
+    @ViewBuilder var incrementButton: StepperButton {
```
sjavora, line 61: `decrementButton` and `incrementButton` are so similar that I keep trying to find a way to make just one declaration with a few alterations 😅 

- `isEnabled` should use environment; the checks around min/max values could be unified (possibly even using a range)
- `value` changes could use a ternary operator

Not sure what to do about `isIncreased` though 🤔

Sources/Orbit/Components/Stepper.swift
```diff
@@ -0,0 +1,262 @@
...
+    let style: Style
+
+    @Environment(\.colorScheme) var colorScheme
+    
+    @State var isIncreased = false
+
+    public var body: some View {
+        HStack(alignment: .center, spacing: 0) {
+            decrementButton
+                .accessibility(addTraits: [.isButton])
```
PavelHolec, line 22: The `StepperButton` is probably already a button (if not, it should be).

We will also need to add specific accessibility identifiers (for UI tests) for all three parts (buttons and a value) so a call site can extract each part.

Sources/Orbit/Components/Stepper.swift
```diff
@@ -0,0 +1,262 @@
...
+        let isEnabled: Bool
+        let style: Stepper.Style
+        let clickAction: () -> Void
+
+        var body: some View {
+            SwiftUI.Button(action: clickAction) {
+                ZStack {
+                    Circle()
+                        .fill(isEnabled ? style.backgroundActiveColor : style.backgroundColor)
+                        .frame(width: 24, height: 24)
```
PavelHolec, line 107: Sizes should not be fixed, but driven by text sizes. Example: Badge or BadgeList icons

Sources/Orbit/Components/Stepper.swift
```diff
@@ -0,0 +1,262 @@
...
+                    icon
+                }
+            }.disabled(!isEnabled)
```
PavelHolec, line 110: A modifier should be on a separate line

PavelHolec:
The animation looks broken/distracting. 
Maybe for now, we can remove the animation until designers request a specific one.


https://user-images.githubusercontent.com/35844477/216377236-c9a9b445-3aeb-4a0e-b2f2-9a3f629a5e5d.mov

PavelHolec:
The storybook icon is similar to the `Select` component. Alternative icon - one of these?

image

Sources/Orbit/Components/Stepper.swift
```diff
@@ -0,0 +1,262 @@
...
+            valueContainer
+                .frame(minWidth: 40)
+                .accessibility(value: .init("\(value)"))
+
+            incrementButton
+                .accessibility(addTraits: [.isButton])
+        }
+    }
+
+    @ViewBuilder var valueContainer: some View {
```
PavelHolec, line 33: Suggestion: `value: some View`

Sources/Orbit/Components/Stepper.swift
```diff
@@ -0,0 +1,262 @@
...
+        case secondary
+
+        public var foregroundColor: Color {
+            switch self {
+                case .primary:                  return .white
+                case .secondary:                return .inkDark.opacity(0.5)
+            }
+        }
+
+        public var foregroundActiveColor: Color {
```
PavelHolec, line 135: The active colors seem off

image

hrach:
> The animation looks broken/distracting. 
> Maybe for now, we can remove the animation until designers request a specific one.
> 
> 
> https://user-images.githubusercontent.com/35844477/216377236-c9a9b445-3aeb-4a0e-b2f2-9a3f629a5e5d.mov
> 
> 

On Android we do animation of the leaving number as well and we animate it only the half of the text height. +fade.

PavelHolec:
> > The animation looks broken/distracting.
> > Maybe for now, we can remove the animation until designers request a specific one.
> > 
> >   
> >     
> >     
> > 
> >     Screen.Recording.2023-02-02.at.17.03.39.mov
> 
> On Android we do animation of the leaving number as well and we animate it only the half of the text height. +fade.

Out of curiosity, when switching between negative numbers - does a minus sign stay in place (not animated) or does every number contains the minus sign?

hrach:
It does not stay

PetroShylaiev:
> The animation looks broken/distracting. Maybe for now, we can remove the animation until designers request a specific one.
> 
>  Screen.Recording.2023-02-02.at.17.03.39.mov

Yeah, to be honest that was the closet variant to match android design. 
The other solution to have it very smooth is using the List with values with a gradient mask applied above so we actually scroll the values like in iOS pickers.

PavelHolec:
> > The animation looks broken/distracting. Maybe for now, we can remove the animation until designers request a specific one.
> > Screen.Recording.2023-02-02.at.17.03.39.mov
> 
> Yeah, to be honest that was the closet variant to match android design. The other solution to have it very smooth is using the List with values with a gradient mask applied above so we actually scroll the values like in iOS pickers.

To not block the merge, we can remove the animation for now and create a follow-up that will:
- add the android-like animation
- make `-` sign static (it looks weird if it moves with numbers)
- disable animation when `accessibilityReduceMotion` is true

sjavora:
> disable animation when accessibilityReduceMotion is true

Maybe just an observation - even for default Apple code, that setting doesn't disable all animations - navigation push for example is the same regardless of whether it is on or off. Maybe we don't need to turn off every single animation 🙂

Sources/Orbit/Components/Stepper.swift
PetroShylaiev: Handled

Sources/Orbit/Components/Stepper.swift
PetroShylaiev: Actually not only the icon we need, we also need appropriate colour for it set. The logic for it is encapsulated into `StepperButton` and is dependant on both `isIncrement` and `isEnabled`. We can move it out, but in that case the sense of `StepperButton` seems to be lost

PavelHolec:
> > disable animation when accessibilityReduceMotion is true
> 
> Maybe just an observation - even for default Apple code, that setting doesn't disable all animations - navigation push for example is the same regardless of whether it is on or off. Maybe we don't need to turn off every single animation 🙂

Could be true. Let's do better than them then 👍

Sources/Orbit/Components/Stepper.swift
```diff
@@ -0,0 +1,262 @@
...
+
+// MARK: - Subtypes
+extension Stepper {
+
+    struct StepperButton: View {
+
+        let isIncrement: Bool
+        let isEnabled: Bool
+        let style: Stepper.Style
+        let clickAction: () -> Void
```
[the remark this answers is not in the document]
PetroShylaiev: Renamed

Sources/Orbit/Components/Stepper.swift
PetroShylaiev: We don't need `isIncreased` anymore since we got rid of animation

Sources/Orbit/Components/Stepper.swift
PetroShylaiev: Added specific identifiers

Sources/Orbit/Components/Stepper.swift
PetroShylaiev: Revised

Sources/Orbit/Components/Stepper.swift
PetroShylaiev: Done

Sources/Orbit/Components/Stepper.swift
```diff
@@ -0,0 +1,262 @@
...
+                .accessibility(value: .init("\(value)"))
+
+            incrementButton
+                .accessibility(addTraits: [.isButton])
+        }
+    }
+
+    @ViewBuilder var valueContainer: some View {
+        Text("\(value)")
+            .foregroundColor(colorScheme == .light ? .cloudDark : .cloudLight)
```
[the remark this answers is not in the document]
PetroShylaiev: Correct, removed

Sources/Orbit/Components/Stepper.swift
PetroShylaiev: Should be better now, pls recheck, the off state now differs from active

hrach:
There was an exciting bug in Compose - disabled animations were disabling cursor blinking in the text field in Compose. It got fixed recently :D

Sources/Orbit/Components/Stepper.swift
```diff
@@ -0,0 +1,265 @@
...
+            case .secondary:                return .cloudNormal
+            }
+        }
+    }
+}
+
+// MARK: - Identifiers
+public extension AccessibilityID {
+    
+    static let stepperActionIncrement         = Self(rawValue: "orbit.stepper.increment")
```
PavelHolec, line 143: There is no reason for names to be different. Should be `let stepperIncrement` as in other components (removing only the `orbit` prefix).

Sources/Orbit/Components/Stepper.swift
PavelHolec: It still looks the same. Can you send a screenshot of how it should look?

PetroShylaiev:
> The storybook icon is similar to the `Select` component. Alternative icon - one of these?
> 
> image

Replaced with "plusminus.circle"

Sources/Orbit/Components/Stepper.swift
PetroShylaiev: **Disabled** state example from Preview
Screenshot 2023-02-09 at 13 56 34

**Disabled** state from Figma
Screenshot 2023-02-09 at 13 57 37

**Selected** state example from Preview
Screenshot 2023-02-09 at 13 55 57

Cannot find the **Selected** state in Figma, is it supposed to be a "Focus" one? P.S. I believe not
Screenshot 2023-02-09 at 14 05 00

PavelHolec: It should be, because we currently do not support hover states (they would only be visible on iPad using pointing device), so the focus state is the one corresponding to "pressed"

PetroShylaiev: Add the pressed state:
Screenshot 2023-02-10 at 16 20 10
Screenshot 2023-02-10 at 16 20 13

Sources/Orbit/Components/Stepper.swift
PetroShylaiev: Renamed

Sources/Orbit/Components/Stepper.swift
```diff
@@ -0,0 +1,262 @@
...
+        self.maxValue = maxValue
+        self.isEnabled = isEnabled
+        self.style = style
+    }
+}
+
+// MARK: - Subtypes
+extension Stepper {
+
+    struct StepperButton: View {
```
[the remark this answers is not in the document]
PetroShylaiev: Handled

Sources/Orbit/Components/Stepper.swift
```diff
@@ -0,0 +1,284 @@
...
+        style: Style
+    ) {
+        self._value = value
+        self.minValue = minValue
+        self.maxValue = maxValue
+        self.style = style
+    }
+}
+
+struct StepperButton: View {
```
PavelHolec, line 70: The `StepperButton` can be extracted in its own file in a `support/components` folder, where we store all these supporting components. It should also be `public` so a call site can use it as a standalone component.

PetroShylaiev: Extracted 👍🏻

PavelHolec (approved):

sjavora:
@PetroShylaiev  So `check-docc-referenced-types` failed - that means that Stepper or some other type from this MR is visible on the front page of the package's documentation (you can find out locally by building documentation in Xcode) - this is not something we want - depending on what types you see there at the bottom, we can figure out where they should go.

Sources/Orbit/Components/Stepper.swift
```diff
@@ -0,0 +1,239 @@
...
+        }
+    }
+}
+
+// MARK: - Identifiers
+public extension AccessibilityID {
+    
+    static let stepperIncrement         = Self(rawValue: "orbit.stepper.increment")
+    static let stepperDecrement         = Self(rawValue: "orbit.stepper.decrement")
+    static let stepperValue                   = Self(rawValue: "orbit.stepper.value")
```
Parabak, line 119: nitpick: wrong padding;

Sources/Orbit/Components/Stepper.swift
```diff
@@ -0,0 +1,239 @@
...
+            valueText
+                .frame(minWidth: 40)
+            
+            incrementButton
+        }
+    }
+    
+    @ViewBuilder var valueText: some View {
+        Text("\(value)")
+            .accessibility(.stepperValue)
```
Parabak, line 29: `.accessibilityValue(.init(value.description))`

Parabak (changes requested):
Pls update accessibility

Sources/Orbit/Components/Stepper.swift
PetroShylaiev: Fixed padding

Sources/Orbit/Components/Stepper.swift
PetroShylaiev: Corrected

sjavora (approved):

Parabak (approved):

PR #465, opened 2023-02-04
Use `Binding`s for control selection values

### LINKED ISSUE #399: Enhance relevant inits with binding parameter
Some components might need to use both initializers:
- with a simple property and callback (when the selection needs to be injected with custom logic)
- with a binding (when the selection should match 1:1 in both directions)

Example is the native `DisclosureGroup` component that as an additional initializer that takes binding: 

> This initializer also let you set or read the expanded state programmatically.

Components:
- `Tag`
- `Collapse`
- `Checkbox`
- `Radio`
- `Switch`
- TBD...

sjavora:
Closes #399.

### DIFF
--- a/Sources/Orbit/Components/Collapse.swift
+++ b/Sources/Orbit/Components/Collapse.swift
@@ -5,13 +5,13 @@ import SwiftUI
 /// - Note: [Orbit definition](https://orbit.kiwi/components/collapse/)
 public struct Collapse<Header: View, Content: View>: View {
 
-    struct _Collapse: View {
-        let header: Header
-        let headerVerticalPadding: CGFloat
-        let content: Content
-        @Binding var isExpanded: Bool
+    let header: Header
+    let headerVerticalPadding: CGFloat
+    let content: Content
+    let isExpanded: Binding<Bool>?
 
-        var body: some View {
+    public var body: some View {
+        BindingSource(isExpanded, fallbackInitialValue: false) { $isExpanded in
             VStack(alignment: .leading, spacing: 0) {
                 SwiftUI.Button {
                     withAnimation(.easeInOut(duration: 0.2)) {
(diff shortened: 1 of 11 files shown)


Sources/Orbit/Support/BindingWrapper.swift
```diff
@@ -0,0 +1,30 @@
...
+
+/// A wrapper that provides a binding to its content.
+///
+/// This binding can either be supplied, in which case it is used directly,
+/// or one is derived from internal state (starting with `defaultValue`.
+///
+/// This is is useful for components that can manage their own state,
+/// but we also want to make it possible for that state to be driven
+/// from the outside if a binding is passed.
+struct BindingWrapper<Value, Content: View>: View {
```
PavelHolec, line 11: Our other wrappers are public - should we expose this one too? If yes, it would need to be added to docs ignore section

PavelHolec: Now I see this one is meant for something else than previews. 

Not sure if it should have a more distinct name, as now it looks almost like a duplicate of `StateWrapper`.

Sources/Orbit/Components/Collapse.swift
```diff
@@ -5,13 +5,13 @@ import SwiftUI
...
-        let content: Content
-        @Binding var isExpanded: Bool
+    let header: Header
+    let headerVerticalPadding: CGFloat
+    let content: Content
+    let isExpanded: Binding<Bool>?
 
-        var body: some View {
+    public var body: some View {
+        BindingWrapper(isExpanded, defaultValue: false) { $isExpanded in
```
PavelHolec, line 14: Is it needed? We can probably now have a `var isExpanded: Binding<Bool>` and set it to dummy `get/set` in the init.

sjavora: I guess I can revert the changes here, but we'd be back to the inner type...

PavelHolec: I will check if what I suggest is technically possible

sjavora: I'd expect it is - just copy the relevant parts of `BindingWrapper` here.

To me the question is more whether there will be more components like `Collapse` here - where the binding is optional and state can be managed internally. Because if there will be more such views, it makes sense to me to extract this concept into a separate type - like `BindingWrapper`.

PavelHolec: The proposed `BindingSource`, `OptionalBinding` are fine. Or maybe something like `BindingResolver`, `BindingSwitcher` ?

PavelHolec (approved):
Nice 👍

Issue #435, opened 2023-01-17
Update `Tabs` and `SegmentedSwitch` components to accept `Hashable` selection


PavelHolec:
The init should be able to take `selection: Hashable` on top of `selectedIndex: Int`.

Make `SegmentedSwitch` `label` parameter optional

PR #484, opened 2023-03-14
Update `SegmentedSwitch` to take a `Hashable` binding

sjavora:
Part of #435.

Turns out updating `Tabs` is nowhere near as straightforward as this, but there is no reason not to update at least one of the components.

The issue should be updated when this is merged.

I did not retain the original `Int` binding, because `Int` is also `Hashable`. The new interface is more in line with SwitUI's [`Picker`](https://developer.apple.com/documentation/swiftui/picker).

Now we can do:
```swift
SegmentedSwitch("Gender", selection: $gender) {
    Text("Male")
        .identifier(Gender.male)

    Text("Female")
        .identifier(Gender.female)
}
```
... but also
```swift
SegmentedSwitch("Gender", selection: $gender) {
    Icon(.genderMan)
        .identifier(Gender.male)

    Icon(.genderWoman)
        .identifier(Gender.female)
}
```
Sky is the limit!

To support the custom builder, I needed to introduce two utility types - `PreferenceReader` and `EnvironmentReader`, which do exactly what you'd expect. There is no way to pass data directly to the builder functions, so this is a workaround.

Most of the implementation should be usable as-is for `Tabs`.

Screenshot 2023-05-19 at 13 50 06
Screenshot 2023-05-19 at 14 08 17
Screenshot 2023-05-19 at 13 52 57


Sources/Orbit/Components/SegmentedSwitch.swift
```diff
@@ -175,10 +188,12 @@ struct SegmentedSwitchPreviews: PreviewProvider {
...
                     "Gender\nmultiline",
-                    firstOption: "Male with long name",
-                    secondOption: "Female with much longer name",
-                    selectedIndex: value
-                )
+                    firstOption: 1,
+                    secondOption: 2,
+                    selection: value
+                ) {
+                    $0 == 1 ? "Male with long name" : "Female with much longer name"
```
PavelHolec, line 195: This way of providing content seems weird. Do we use it somewhere else? 

Maybe we could introduce custom builders - that variant would look like this:

```
SegmentedSwitch("Gender", selectedValue: 2) {
   Segment(1, "Male")
   Segment(2, "Female")
}
```

sjavora: The way `Picker` does it is by assigning `.tag`s to the individual segments. Since the primary way we want to use this is not to specify custom views, but only strings (while retaining a way to set custom views), I figured this made more sense. Usually you would use an enum, not `Int`, so you would have a `switch` here.

I actually tried what you propose when I attempted to update `Tabs`, but the types aren't inferred in the builder for some reason, so you can't just use the `.caseName` notation.

Since this is limited to two options, another way to do it would be to pass the strings directly as parameters, eliminating this closure.

Sources/Orbit/Components/SegmentedSwitch.swift
```diff
@@ -126,8 +100,58 @@ public extension AccessibilityID {
...
+    static let verticalTextPadding: CGFloat = .small // = 44 @ normal text size
+
+    public static func buildBlock(_ first: some View, _ second: some View) -> some View {
+        HStack(spacing: 0) {
+            item {
+                first
+                    .accessibility(.segmentedSwitchFirstOption)
+            }
+
+            TextStrut()
```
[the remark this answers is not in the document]
sjavora: Updated.

PavelHolec (approved):
Very nice

PR #485, opened 2023-03-16
Add tint option to `BarButton`

PavelHolec:
Fixed button highlight without having to specify hightlight color.

https://user-images.githubusercontent.com/35844477/225601957-de96a243-2c8a-4422-8d50-35ed317a7873.mov


Sources/Orbit/Support/Components/BarButton.swift
```diff
@@ -28,9 +29,16 @@ public struct BarButton: View {
         }
     }
 
-    public init(_ content: Icon.Content, size: Icon.Size = .large, alignment: TextAlignment = .center, action: @escaping () -> Void) {
+    public init(
+        _ content: Icon.Content,
+        size: Icon.Size = .large,
+        tint: Color? = nil,
+        alignment: TextAlignment = .center,
+        action: @escaping () -> Void = {}
```
sjavora, line 37: Didn't we stop giving actions default values?

Sources/Orbit/Support/Components/BarButton.swift
```diff
@@ -28,9 +29,16 @@ public struct BarButton: View {
         }
     }
 
-    public init(_ content: Icon.Content, size: Icon.Size = .large, alignment: TextAlignment = .center, action: @escaping () -> Void) {
+    public init(
+        _ content: Icon.Content,
+        size: Icon.Size = .large,
+        tint: Color? = nil,
```
sjavora, line 35: We can't get tint from environment, right?

PavelHolec: It is stupid, but sadly not.

https://kylehugh.es/documents/how-to-read-foreground-and-tint-colors-in-swiftui-views/

sjavora (approved):

PR #489, opened 2023-03-20
Add accessibilityValue to `Stepper`

### LINKED ISSUE #488: Stepper accessibility value cannot be read

sjavora:
Resolves #488


Sources/Orbit/Components/Stepper.swift
```diff
@@ -25,8 +25,14 @@ public struct Stepper: View {
     }
     
     @ViewBuilder var valueText: some View {
-        Text("\(value)")
-            .accessibility(.stepperValue)
+        if #available(iOS 14, *) {
```
PavelHolec, line 28: We should be able to use the iOS13 version `accessibility(value: Text)` as we do it for other components.

PetroShylaiev (approved):

Sources/Orbit/Components/Stepper.swift
sjavora: You're right, it works with that, too. Replaced.

PavelHolec (approved):

PR #495, opened 2023-03-22
Update `StepperButton` size

PavelHolec:
Updated the button size to match design

image

#### XXL

image


PetroShylaiev:
In case of the value equals “10”, will the distance between buttons stay the same as with the value “9”?

PetroShylaiev (approved):

sjavora (approved):

PavelHolec:
> In case of the value equals “10”, will the distance between buttons stay the same as with the value “9”?
> 
> UPD: checked the branch, with 2 digits it works fine, with 3 digits the distance increases, but that should be ok

The minimalWidth was also taken from figma to match the design.

image

PR #498, opened 2023-03-22
Update `HorizontalScroll` component with snapping support

PavelHolec:
Programatic scrolling is implemented as a iOS14+ feature.

- Removed the dual native/custom scrollview
- Programatic scrolling and/or snapping can work together.

Contains `identifier()` wrapper over native `id()`.

image

https://user-images.githubusercontent.com/35844477/227236022-638d48b8-2795-4af1-8fdd-a172fbcef458.mov


Sources/Orbit/Components/HorizontalScroll.swift
```diff
@@ -23,301 +28,521 @@ public enum HorizontalScrollItemHeight {
...
+    @ViewBuilder var heightReader: some View {
+        GeometryReader { geometry in
+            Color.clear.preference(
+                key: ContentSizePreferenceKey.self,
+                value: geometry.size
+            )
+        }
+    }
+
+    @ViewBuilder var scrollViewWidthReader: some View {
```
sjavora, line 142: Why is one just `heightReader` and the other `scrollViewWidthReader`?

Sources/Orbit/Components/HorizontalScroll.swift
```diff
@@ -23,301 +28,521 @@ public enum HorizontalScrollItemHeight {
...
     ///   - spacing: Spacing between items. Default value is `.small`.
     ///   - itemWidth: Horizontal sizing style of all items. Default value is `.ratio()`.
     ///   - itemHeight: Horizontal sizing style of all items. Default value is `.intrinsic`.
     ///   - maxItemWidth: Maximal item width that caps the result of `itemWidth`.
     ///                   Default value is derived from ``Layout/readableMaxWidth``.
-    ///   - minHeight: Minimal height of component. Default value is `nil`.
-    ///   - horizontalPadding: Horizontal padding inside the ScrollView. Default value is `.xSmall`.
-    ///   - verticalPadding: Horizontal padding inside the ScrollView. Can be used to fix overlay issues. Default value is `.xSmall`.
     ///   - content: Child items that will be horizontally laid out based on above parameters.
+    /// - Warning: when using snapping effect than programatic scrolling using `ScrollViewReader`
```
sjavora, line 214: then

Sources/Orbit/Components/HorizontalScroll.swift
```diff
@@ -23,301 +28,521 @@ public enum HorizontalScrollItemHeight {
...
+            case .intrinsic:                    self.snapping = false
+            case .ratio(_, let snapping):       self.snapping = snapping
+            case .custom(_, let snapping):      self.snapping = snapping
         }
     }
-}
 
-// MARK: - Previews
-struct HorizontalScrollPreviews: PreviewProvider {
+    private func scrollToID(id: AnyHashable?, geometry: GeometryProxy) {
```
sjavora, line 236: Can `id` be `nil`?

Maybe rename to `scroll(to id:`

Sources/Orbit/Components/HorizontalScroll.swift
```diff
@@ -23,301 +28,521 @@ public enum HorizontalScrollItemHeight {
...
+        guard let id else { return }
 
-    static var previews: some View {
-        PreviewWrapper {
-            content
+        if let itemToScrollTo = idPreferences.first(where: { $0.id == id }) {
```
sjavora, line 239: Could be combined with the `guard`.

Sources/Orbit/Components/HorizontalScroll.swift
```diff
@@ -23,301 +28,521 @@ public enum HorizontalScrollItemHeight {
...
-                Color.greenLight.frame(height: 30)
-                Color.greenLight.frame(height: 50)
-                Color.greenLight.frame(height: 20)
-            }
-            .border(Color.cloudDark)
+        guard isContentBiggerThanScrollView else { return }
+
+        if currentOffset > scrollViewWidth / 5 {
+            stopSnappingAnimation()
+        } else  if currentOffset < maxOffset - scrollViewWidth / 5 {
```
sjavora, line 266: extra space

Sources/Orbit/Components/HorizontalScroll.swift
```diff
@@ -23,301 +28,521 @@ public enum HorizontalScrollItemHeight {
...
+
+        snappingOffset = scrollSnappingOffset(currentOffset: currentOffset)
+    }
+
+    private func gestureEnded(gesture: DragGesture.Value) {
+        scrollingInProgress = false
+
+        guard let itemWidth = resolvedItemWidth else { return }
+
+        guard isContentBiggerThanScrollView else { return }
```
sjavora, line 311: `guard`s could be combined.

Sources/Orbit/Components/HorizontalScroll.swift
sjavora, line 239: It seems like the order of `idPreferences` doesn't matter. Since they are `Hashable`, why not make it a `Set`?

PavelHolec: The updated version now counts with the proper order.

PavelHolec: The implementation changed a bit, should be updated.

Sources/Orbit/Components/HorizontalScroll.swift
PavelHolec: The implementation changed a bit, should be updated.

Sources/Orbit/Components/HorizontalScroll.swift
PavelHolec: Renamed to `contentHeightReader` and `availableWidthReader`

Sources/Orbit/Components/HorizontalScroll.swift
PavelHolec: Renamed to match the proxy name as `scrollTo(:)`

Sources/Orbit/Support/Layout/HorizontalScrollReader.swift
```diff
@@ -0,0 +1,110 @@
+import SwiftUI
+
+@available(iOS 14, *)
+public class HorizontalScrollViewProxy: ObservableObject {
+
+    @Published fileprivate var scrollTarget: ScrollTargetValue?
+
+    public func scrollTo<ID>(_ id: ID, animated: Bool = true) where ID : Hashable {
```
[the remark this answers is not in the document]
sjavora: Ok, though `some Hashable` can still be done. Maybe even on the other method, since `AnyHashable` is also `Hashable`...

Sources/Orbit/Components/HorizontalScroll.swift
```diff
@@ -1,19 +1,11 @@
 import SwiftUI
+import Combine
 
 public enum HorizontalScrollItemWidth {
-    /// Each item will use its own intrinsic width.
-    case intrinsic
-    /// Width ratio compared to container width.
-    case ratio(CGFloat = 0.8)
+    /// Width ratio calculated from the available container width.
+    case ratio(CGFloat = 0.48, maxWidth: CGFloat? = Layout.readableMaxWidth - 270)
```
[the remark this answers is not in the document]
PavelHolec: The maxWidth was already there, just have been moved here. It is there as a cap for regular horizontal presentation (so that the call site does not have to switch over it and keep the same ratio).

The new 0.48 is a ratio that fits 2 items and makes a 3rd item edge visible. This works as a better defaults for both compact and regular widths.

For example, a ratio of 0.5 or 1.0 would fit the items exactly, but the edges of next scrollable items would not be visible.

Sources/Orbit/Support/Layout/HorizontalScrollReader.swift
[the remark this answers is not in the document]
PavelHolec: Updated this one to use `some Hashable`, but kept it in the method inside the other component - otherwise `as AnyHashable` would need to be cast in the `first()` method.

sjavora (approved):

PR #499, opened 2023-03-23
Update `Dialog` component

PavelHolec:


Sources/OrbitStorybook/Detail/Items/StorybookDialog.swift
```diff
@@ -72,6 +62,7 @@ struct StorybookDialogPreviews: PreviewProvider {
     static var previews: some View {
         OrbitPreviewWrapper {
             StorybookDialog.basic
+            StorybookDialog.normal
```
sjavora, line 65: `basic` **and** `normal`? 🤔

Sources/Orbit/Components/Dialog.swift
```diff
@@ -77,21 +78,43 @@ public struct Dialog: View {
...
+
+    /// Creates Orbit Dialog component.
+    public init(
+        illustration: Illustration.Image = .none,
+        title: String = "",
+        description: String = "",
+        style: Style = .primary,
+        buttons: Buttons,
+        descriptionLinkAction: @escaping TextLink.Action = { _, _ in }
+    ) where Content == EmptyView {
```
sjavora, line 108: Why not set the default content value in the other init?

```swift
@ViewBuilder content: () -> Content = { EmptyView() }
```

That way, this init wouldn't have to be written.

PavelHolec: Forgot we can finally do that

Sources/OrbitStorybook/Detail/Items/StorybookDialog.swift
PavelHolec: This is using the existing naming (basic stands for the tab content and the normal is used within the component preview)

sjavora (approved):

PR #522, opened 2023-04-04
Support native modifiers on `Text`

PavelHolec:
Matching the SwiftUI API: 
- richer text concatenation (this will later allow new inits that take `Text` on top of `String`)
- support for subview modifiers (iOS16+ mostly)
- simplify the `Text` init (keeping only what is defined by Orbit - `size`)

image


Sources/Orbit/Support/Text/TextBuildable.swift
```diff
@@ -1,41 +1,331 @@
 import SwiftUI
 
+// Modifiers that shadow the SwiftUI modifiers need to have the exact signature
+// in order to prefer the method of the Text over the generic View version
```
sjavora, line 4: So this exists in order for these modifierrs to return the same type (`Self`)?

PavelHolec: Yes exactly - just as the native `Text` does. Otherwise, it would be `just` a View. This makes it possible to build the `Text` properties and later concatenate them using the same syntax as the native one.

sjavora (approved):

PR #548, opened 2023-04-21
Release fixes for 0.10.0

PavelHolec:


Sources/Orbit/Support/TextFields/TextField.swift
```diff
@@ -200,17 +184,27 @@ public struct TextField: UIViewRepresentable, TextFieldBuildable {
...
                     return false
                 case .reject:
                     return false
             }
         }
+
+        func replaceValue(with text: String) {
+            textFieldValue = text
+
+            DispatchQueue.main.async { [weak self] in
```
username0x0a, line 203: Main thread dispatch isn't retained anywhere so it should be safe to go with strong reference. 👍

sjavora: `Task { @MainActor` is also an option...

PavelHolec: Updated with `Task { @MainActor` in 3 places where we need to dispatch the update to prevent the realtime SwiftUI warning "Modifying state during view update, this will cause undefined behavior"

username0x0a (approved):

PR #571, opened 2023-05-19
Configurable haptics

### LINKED ISSUE #130: Haptics for interactive components should be configurable
Haptics {
   case none
   case `default`
   case custom(...)
}

Each component will use the .default value as default, but the actual intensity/type value will be up to component.

sjavora:
Closes #130.

### DIFF
--- a/Sources/Orbit/Support/Environment Keys/HapticsEnabledKey.swift
+++ b/Sources/Orbit/Support/Environment Keys/HapticsEnabledKey.swift
@@ -1,0 +1,22 @@
+import SwiftUI
+
+struct HapticsEnabledKey: EnvironmentKey {
+    static let defaultValue = true
+}
+
+public extension EnvironmentValues {
+
+    /// Whether haptic feedback on controls is enabled.
+    var isHapticsEnabled: Bool {
+        get { self[HapticsEnabledKey.self] }
+        set { self[HapticsEnabledKey.self] = newValue }
+    }
+}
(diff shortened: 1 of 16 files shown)


Sources/Orbit/Support/Environment Keys/HapticsEnabledKey.swift
```diff
@@ -0,0 +1,22 @@
...
+    var isHapticsEnabled: Bool {
+        get { self[HapticsEnabledKey.self] }
+        set { self[HapticsEnabledKey.self] = newValue }
+    }
+}
+
+public extension View {
+
+    /// Disables haptic feedback on controls.
+    func disabledHaptics(_ disabled: Bool = true) -> some View {
```
PavelHolec, line 19: Suggestion - the Apple API seems to have switched naming for such cases. We could go with `hapticsDisabled()`

image

PavelHolec (approved):
Nice. I was also thinking we could also add a `HapticsProvider.disable()` to turn off the feedback completely, if a call site for some reason wants to disable our feedback implementation completely. Not necessary though.

Sources/Orbit/Support/Environment Keys/HapticsEnabledKey.swift
sjavora: Updated.

PavelHolec (approved):

PR #576, opened 2023-05-26
Simplify `SegmentedSwitch` implementation

### LINKED ISSUE #575: Simplify `SegmentedSwitch` implementation
`SegmentedSwitch` can be simplified by removing the custom builder (and the utility views added because of it) and completely separating the input handling from the visual.

sjavora:
No changes in functionality, just code simplification. No more custom builder, so there is no longer a two-view limit, but everything else should stay the same - and it seems this will work with a `ForEach` as well, so this is a good preparation for `Tabs`.

Closes #575.


Sources/Orbit/Components/SegmentedSwitch.swift
```diff
@@ -19,42 +21,54 @@ public struct SegmentedSwitch<Selection: Hashable, Content: View>: View {
...
-                    .environment(\.segmentSelection, selectionBinding)
-                    .overlay(separator)
-                    .backgroundPreferenceValue(IDPreferenceKey.self) { preference in
-                        selectionBackground(preference)
-                    }
-                    .fixedSize(horizontal: false, vertical: true)
+                HStack(spacing: 0) {
+                    content
+                        .allowsHitTesting(false)
+                        .frame(maxWidth: .infinity)
```
PavelHolec, line 27: We could add a support for `idealSize` here, as we support it in other components where it makes sense. Here, call site might want to have the switch that only takes a necessary width.

Sources/Orbit/Components/SegmentedSwitch.swift
```diff
@@ -19,42 +21,54 @@ public struct SegmentedSwitch<Selection: Hashable, Content: View>: View {
...
-            .frame(width: borderWidth)
-            .frame(maxHeight: .infinity)
-            .padding(.vertical, BorderWidth.active)
+    @ViewBuilder func interactionBackground(_ preferences: [IDPreference]) -> some View {
+        GeometryReader { geometry in
+            ForEach(Array(preferences.enumerated()), id: \.element.id) { index, preference in
+                Color.clear
+                    .contentShape(Rectangle())
+                    .segmentFrame(bounds: preference.bounds, in: geometry)
+                    .onTapGesture {
```
PavelHolec, line 51: I thought we would go with using native Button, as it gives us "touchDown" highlight support and proper native accessibility. (The `isSelected` trait would still be kept, but the rest would be for free)

Sources/Orbit/Components/SegmentedSwitch.swift
```diff
@@ -19,42 +21,54 @@ public struct SegmentedSwitch<Selection: Hashable, Content: View>: View {
...
-                    let itemWidth = geometry.size.width / 2
-
                     background
-                        .frame(width: itemWidth)
-                        .offset(x: (geometry[anchor].minX / itemWidth).rounded(.down) * itemWidth)
+                        .segmentFrame(bounds: bounds, in: geometry)
                 }
             } else {
                 background
+                    .overlay(separator)
```
PavelHolec, line 71: To be more precise, the HStack that hosts the content should use `spacing` in the same size of separator, so that the center separator overlay fits exactly between expected two segments.

If a call-site provides anything else than two views, the overlay would be broken. The Tabs-like solution probably can solve this in a follow-up.

Sources/Orbit/Components/SegmentedSwitch.swift
```diff
@@ -95,65 +109,30 @@ public struct SegmentedSwitch<Selection: Hashable, Content: View>: View {
...
-    static let defaultValue: Binding<AnyHashable?> = .constant(nil)
-}
-
-extension EnvironmentValues {
-    var segmentSelection: Binding<AnyHashable?> {
-        get { self[SegmentSelectionKey.self] }
-        set { self[SegmentSelectionKey.self] = newValue }
-    }
+public extension AccessibilityID {
+    static let segmentedSwitchFirstOption = Self(rawValue: "orbit.segmentedswitch.first")
```
PavelHolec, line 124: These are now unused. With any content, call site needs to provide its own identifiers.

PavelHolec (commented):
👍

Sources/Orbit/Components/SegmentedSwitch.swift
sjavora: But that highlight only applies to the button's title, not whatever is under it... at least from experimentation. I suppose we could create a custom button style, but the default one doesn't seem to work the way you think.

sjavora: So it's kind of possible to do this with an overlay in a custom button style.

Do we want the overlay to also "dim" the content...
https://github.com/kiwicom/orbit-swiftui/assets/12349477/5391ceef-3f96-4a45-8114-4e8bd3f9bd81

... or just the background
https://github.com/kiwicom/orbit-swiftui/assets/12349477/739185c4-d2ae-4502-bf21-faaddc6a7692
?

(Showing this here on `Tabs`, but the principle is the same)

PavelHolec: Yep, what I meant - a custom `ButtonStyle` that mimics the active/inactive background of what is under the title

True, the title itself would normally stay as-is with this approach, but in this case, that is probably what we want - to only decorate the background of the buttonm not the title itself. If we wanted to do that though, it would be much more difficult (we would again need to pass some `isSelected/isActive`  environment to the active component)

PavelHolec: Now we are saying: take the ViewBuilder content, obtain how many items are actually there and generate as many buttons based on their sizes (+ padding to each one). If we are fine with button background changes only, this should be all we need.

sjavora:
@PavelHolec updated everything, including the button background highlight - I just used what looked right to me, so this will probably need to be tweaked. I think we'll want to use the same thing for `Tabs`, so I extracted the style to a separate file.

Sources/Orbit/Components/SegmentedSwitch.swift
PavelHolec: The identifier can be checked in SwiftUI previews as well

image

sjavora:
@PavelHolec please recheck - probably the whole thing. Buttons are now responsible for drawing the selected segment, so the custom style can shrink the backgournd kind of like the official component. I tried to update the fade/highlight on the unselected segments, but of course we're limited here by the fact that the buttons don't actually contain the content, but are drawn under them, so no fade of the titles and no shrinking of the titles, either.

Still, I think this is now pretty damn good, considering how much time (way more than I'd like) I spent on it. See video for what the interactions look like now.

Ideal size is also supported.

https://github.com/kiwicom/orbit-swiftui/assets/12349477/5e711833-0429-4645-9183-de086a3818a4

PavelHolec (approved):

PR #581, opened 2023-05-31
Add `textFontWeight` modifier

### LINKED ISSUE #578: Add `.textFontWeight` modifier
In order to support `Tabs` (but not only for that), we should introduce a modifier to mimic the behavior of the native [fontWeight](https://developer.apple.com/documentation/swiftui/view/fontweight(_:)), which is only available from iOS 16.

sjavora:
Closes #578 , needed for #577.

### DIFF
--- a/Sources/Orbit/Components/Text.swift
+++ b/Sources/Orbit/Components/Text.swift
@@ -13,6 +13,7 @@ public struct Text: View, FormattedTextBuildable {
     @Environment(\.lineSpacing) private var lineSpacing
     @Environment(\.textAccentColor) private var textAccentColor
     @Environment(\.textColor) private var textColor
+    @Environment(\.textFontWeight) private var textFontWeight
     @Environment(\.sizeCategory) private var sizeCategory
 
     private let content: String
@@ -106,8 +107,7 @@ public struct Text: View, FormattedTextBuildable {
         if #available(iOS 16.0, *), let isBold {
             return text
                 .bold(isBold)
-        }
-        else if let isBold, isBold {
+        } else if let isBold, isBold {
             return text
                 .bold()
         } else {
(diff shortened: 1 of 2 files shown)


Sources/Orbit/Support/Environment Keys/TextFontWeightKey.swift
```diff
@@ -0,0 +1,28 @@
...
+
+    /// Set the font weight for any text based Orbit component within the view hierarchy.
+    ///
+    /// - Parameters:
+    ///   - fontWeight: A font weight that will be used in text based Orbit components.
+    ///                 Pass `nil` to ignore environment font weight and to allow the system
+    ///                 or the container to provide its own font weight.
+    ///                 If a container-specific override doesn’t exist, the `regular` will be used.
+    func textFontWeight(_ fontWeight: Font.Weight?) -> some View {
+        environment(\.textFontWeight, fontWeight)
```
PavelHolec, line 26: `@available obsoleted iOS16, renamed: "Use native fontWeight()"` could be added to notify about this being only a workaround for lower iOS versions.

PavelHolec (approved):

Sources/Orbit/Components/Text.swift
```diff
@@ -106,8 +107,7 @@ public struct Text: View, FormattedTextBuildable {
         if #available(iOS 16.0, *), let isBold {
             return text
                 .bold(isBold)
-        }
-        else if let isBold, isBold {
+        } else if let isBold, isBold {
```
username0x0a, line 110: `} else if isBold == true {` directly? 🤔

Also, aren't we piling up unnecessary attributes (`Text(...).bold(false).italic(false)`) or it is built up for a reason to maybe override it in the underlying hierarchy? Not sure whether this is applicable on some concatenation atm.

sjavora: I'm going with "not my problem" on this one... I just saw suboptimal formatting 😅

username0x0a: Legit I guess 😄👌

PR #603, opened 2023-06-20
Add inline label support to `Select`

PavelHolec:
image


Sources/Orbit/Support/Forms/InputContent.swift
```diff
@@ -1,30 +1,35 @@
...
-    @ViewBuilder var suffix: Suffix
+    private let label: String
+    private let state: InputState
+    private let message: Message?
+    private let isPressed: Bool
+    private let isEditing: Bool
+    private let isPlaceholder: Bool
+    @ViewBuilder private let content: Content
+    @ViewBuilder private let prefix: Prefix
+    @ViewBuilder private let suffix: Suffix
```
sjavora, line 20: Should these still be `@ViewBuilder` if they are called in the inits with `()`? By the time they are assigned, they should be resolved values, shouldn't they?

PavelHolec: The `@ViewBuilder` probably does not need to be specified, but it may be better to keep it for semantics (same reason we do it for `@ViewBuilder var` properties and functions) and to match the init parameters.

Sources/Orbit/Support/Forms/InputContent.swift
```diff
@@ -46,7 +51,13 @@ struct InputContent<Content: View, Prefix: View, Suffix: View>: View {
         .overlay(border)
     }
 
-    @ViewBuilder var border: some View {
+     @ViewBuilder private var compactLabel: some View {
+         FieldLabel(label)
+             .textColor(isPlaceholder ? .inkDark : .inkLight)
+             .padding(.leading, prefix.isEmpty ? .small : 0)
```
PavelHolec, line 57: Wrong leading padding of inline label in case of no icon.

PR #608, opened 2023-06-22
Extract `textSize` and `iconSize` from `Text` and `Icon` to modifiers

PavelHolec:


Sources/Orbit/Components/Button.swift
```diff
@@ -30,10 +30,11 @@ public struct Button<LeadingIcon: View, TrailingIcon: View>: View {
                 trailingIcon
             }
         )
+        .textSize(custom: size.textSize.value)
```
sjavora, line 33: Can we put this inside OrbitButtonStyle?

Sources/Orbit/Components/Icon.swift
```diff
@@ -411,17 +423,19 @@ struct IconPreviews: PreviewProvider {
...
 
-                + Icon(.informationCircle, size: .small)
+                + Icon(.informationCircle)
                     .iconColor(.blueNormal)
-                + Icon(.informationCircle, size: .small)
+                + Icon(.informationCircle)
                     .baselineOffset(.xxxSmall)
             )
+            .textSize(.small)
+            .iconSize(.small)
```
sjavora, line 438: Isn't this something you said we don't need on slack?

Sources/Orbit/Support/Environment Keys/TextIsCopyableKey.swift
```diff
@@ -0,0 +1,26 @@
+import SwiftUI
+
+struct TextIsCopyableKey: EnvironmentKey {
+    static let defaultValue = false
+}
+
+public extension EnvironmentValues {
+
+    /// A text copyable status stored in a view’s environment, used for Orbit text components, such as `Text` or `Heading`.
+    var textIsCopyable: Bool {
```
sjavora, line 10: I liked "selectable" better... is "copy" really the only option in all circumstances? Can't the text for example also be read by Siri or something?

Sources/Orbit/Components/Text.swift
```diff
@@ -64,7 +67,7 @@ public struct Text: View, FormattedTextBuildable {
     }
 
     @ViewBuilder private var selectableLabelWrapper: some View {
-        if isSelectable {
+        if textIsCopyable {
```
sjavora, line 70: If you want to stick with copyable, you should probably rename the rest of "selectable" here.

sjavora (changes requested):

Sources/Orbit/Support/Environment Keys/TextIsCopyableKey.swift
PavelHolec: Me too, but SwiftUI already has a `textSelection` modifier that "Use this method to indicate if text is selectable". But this is in regards to Mac selectability of any subtext. 

We are allowing text selection here, we only allow user to "long tap" and "copy" the text.

Sources/Orbit/Components/Text.swift
PavelHolec: Updated

Sources/Orbit/Components/Icon.swift
PavelHolec: Updated

Sources/Orbit/Components/Button.swift
PavelHolec: Updated

sjavora (approved):

PR #612, opened 2023-06-27
`Alert` updates

### LINKED ISSUE #533: Update `Alert` design
- Smaller spacing between buttons (8pt)
- Smaller corner radius for buttons (3pt)
- Use “ButtonLink” for action in Inline variant

The Alert buttons API should be updated as well to remove the awkward enum with associated action handlers

#### Option 1
They may need to be extracted to internal components (`AlertButton`).
The Buttons API could be improved to take a ResultBuilder of Up to these new 2 `AlertButton` components, instead of passing cumbersome parameters.

#### Option 2
The Alert would take action closures for each button and based on their non-nil values, internal buttons would be automatically constructed.

The init taking only one action handler would have `isInline: Bool = false` parameter, while the one with two action handlers would be automatically non-inline.

```
Alert(...) {
    // primary (or single) button action
} secondaryAction: {
    // secondary action
} icon: {
    CountryFlag(..)
}

Alert(..., isInline: true) {
    // primary (single) button action
}
```

For the corner radius of buttons, it might be useful if we fully extract our button as a `buttonStyle` (it partially is) with configurable corner radius (automatic radius is supported only in iOS16+) so that we can use Native `SwiftUI.Button` with this styling override in case of Alert.

Background color from normal colors:
- BG opacity: 12%
- BG opacity: 24%

Text color:
- Default: Dark
- Active: Dark :active

image

sjavora:
Updates the design, splits `Alert` into:
- `Alert`
- `AlertContent`
- `AlertInline`

Additionally introduces a result builder for each alert variant (and `_ConditionalContent` to support those builders).

Resolves #533.

| | `Alert` | `AlertInline` |
|--|--|--|
| | Screenshot 2023-06-27 at 23 13 44 | Screenshot 2023-06-27 at 23 14 01  |
| subtle | Screenshot 2023-06-27 at 23 14 30 | Screenshot 2023-06-27 at 23 14 45 |


Sources/Orbit/Components/AlertInline.swift
```diff
@@ -0,0 +1,216 @@
...
+public enum AlertInlineButtonsBuilder {
+
+    public static func buildBlock(_ emptyView: EmptyView) -> EmptyView {
+        emptyView
+    }
+
+    public static func buildBlock(_ button: some View) -> some View {
+        button
+            .suppressButtonStyle()
+            .buttonStyle(OrbitButtonLinkButtonStyle(type: .status(nil), icon: {}, trailingIcon: {}))
```
PavelHolec, line 69: We could add { EmptyView() } as default values for both `OrbitButtonLinkButtonStyle` and `OrbitButtonStyle`

Sources/Orbit/Components/AlertInline.swift
```diff
@@ -0,0 +1,216 @@
...
+/// Breaks the main user flow to present information.
+///
+/// There are times when just simple information isn’t enough and the user needs
+/// to take additional steps to resolve the problem or get additional details.
+///
+/// - Note: [Orbit definition](https://orbit.kiwi/components/alert/)
+/// - Important: Component expands horizontally unless prevented by `fixedSize` or `idealSize` modifier.
+public struct AlertInline<Icon: View, Button: View>: View {
+
+    let title: String
```
PavelHolec, line 12: We can start adding `private` to all properties to avoid accidental usage of synthesised inits in previews.

Sources/Orbit/Components/AlertInline.swift
```diff
@@ -0,0 +1,216 @@
...
+    let isSubtle: Bool
+    @ViewBuilder let button: Button
+    @ViewBuilder let icon: Icon
+
+    public var body: some View {
+        AlertContent(title: title, description: "", isInline: true, isSubtle: isSubtle) {
```
PavelHolec, line 18: `description` could be made optional

Sources/Orbit/Components/AlertInline.swift
PavelHolec, line 69: We will need a `dark` variant of color for `textColor`. The easiest way is probably to add overridable `textColor` to `OrbitButtonLinkButtonStyle` init. (We probably cannot access environment key from AlertInlineButtonsBuilder to be able to specify `.textColor(status.darkColor)` there - plus we would need to add that override support inside `OrbitCustomButtonContent`)

PavelHolec (approved):

Sources/Orbit/Components/AlertInline.swift
sjavora: These were replaced with `AlertButtonStyle` and `AlertInlineButtonStyle`, which have no icons.

Sources/Orbit/Components/AlertInline.swift
sjavora: Updated.

Sources/Orbit/Components/AlertInline.swift
sjavora: Added.

Sources/Orbit/Components/AlertInline.swift
sjavora: Added custom styles to both alert types.

Sources/Orbit/Components/Alert.swift
```diff
@@ -12,284 +12,195 @@ import SwiftUI
...
-                    .strokeBorder(strokeColor, lineWidth: 1)
-            )
-            .overlay(
-                outlineColor
-                    .frame(height: 3),
-                alignment: .top
-            )
-    }
+// MARK: - Types
+struct AlertButtonStyle: PrimitiveButtonStyle {
```
PavelHolec, line 96: This variant can be made public so the call site can replicate more custom layout using these alert buttons. The inline variant may stay internal or can be made public too.

Sources/Orbit/Support/Alert/AlertContent.swift
```diff
@@ -0,0 +1,187 @@
...
+            title: title,
+            description: description,
+            isInline: true,
+            isSubtle: false
+        ) {
+            contentPlaceholder
+        } icon: {
+            Icon(.informationCircle)
+        } buttons: {
+            Button("Primary") {}
```
PavelHolec, line 182: The preview shows default buttons - maybe intentional, but a `.buttonStyle` could be added to show the example fully customized.

image

Sources/Orbit/Support/Alert/AlertContent.swift
```diff
@@ -0,0 +1,187 @@
+import SwiftUI
+
+struct AlertContent<Content: View, Icon: View, Buttons: View>: View {
```
PavelHolec, line 3: As the `AlertContent` is internal, it could use the new `isSubtle` environment key.

Sources/Orbit/Components/Alert.swift
```diff
@@ -12,284 +12,195 @@ import SwiftUI
...
         }
+        .textFontWeight(.medium)
+        .textSize(.small)
     }
 
-    var strokeColor: Color {
-        switch style {
-            case .status(_, true):              return .cloudLightHover
-            case .status(let status, false):    return (status ?? defaultStatus).lightHoverColor
+    var textColor: Color {
```
PavelHolec, line 124: A `textColorActive` should be also added, using the `DarkActive` variant of color.

PavelHolec: Suggestion - we could replace the textColor parameter in `OrbitCustomButtonContent ` with environment key - after all, the symmetry with active variant is not better than duplicating the property with existing key 🙏

Sources/Orbit/Support/Alert/AlertContent.swift
sjavora: **Any**  button style or the one that should be there?

Sources/Orbit/Components/Alert.swift
PavelHolec: I was thinking of scenario where an Alert should show some custom content "below" buttons and a call site would need to be able to create these buttons for such specific layout. It should not do any harm in exposing it. If not, we can remove the public from one of the `makeBody` methods.

Sources/Orbit/Support/Alert/AlertContent.swift
sjavora: I'll use the "correct" one.

Sources/Orbit/Components/Alert.swift
sjavora: Updated.

Sources/Orbit/Support/Alert/AlertContent.swift
sjavora: Done.

Sources/Orbit/Components/Alert.swift
sjavora: Done.

PR #633, opened 2023-07-19
Improve `EmptyView` detection

PavelHolec:
Not perfect, but fixes release issues related to custom Icon content.

### DIFF
--- a/Sources/Orbit/Components/Badge.swift
+++ b/Sources/Orbit/Components/Badge.swift
@@ -6,7 +6,7 @@ import SwiftUI
 /// They can be updated when a status changes, but they should not be actionable.
 ///
 /// - Note: [Orbit definition](https://orbit.kiwi/components/badge/)
-public struct Badge<LeadingIcon: View, TrailingIcon: View>: View {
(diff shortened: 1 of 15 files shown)


Sources/Orbit/Support/Layout/OptionalEmptyView.swift
```diff
@@ -0,0 +1,5 @@
+
+/// A type that represents views that can optionally result in `EmptyView`.
+protocol OptionalEmptyView {
```
sjavora, line 3: Wouldn't `PossiblyEmptyView` or `PotentiallyEmptyView` be more accurate? `Optional` has a meaning in Swift already...

Also, this is internal, so `///` doesn't make sense IMO.

sjavora (changes requested):

Sources/Orbit/Support/Layout/OptionalEmptyView.swift
PavelHolec: I was deciding between `PotentiallyEmptyView` and this one, will rename it. The comment is there mostly for orbit devs. But the better name may make it less necessary.

sjavora (approved):

PR #653, opened 2023-08-15
Update inline label design

PavelHolec:
image


sjavora (approved):

PavelHolec:
@sjavora had to update the TextField component to respect our text/font modifiers (otherwise there would be a size/baseline mismatch)

Sources/Orbit/Components/InputField.swift
```diff
@@ -117,24 +108,25 @@ public struct InputField<Prefix: View, Suffix: View>: View, TextFieldBuildable {
...
             case .compact:          return label
         }
     }
 
     private var textFieldLeadingPadding: CGFloat {
-        prefix.isEmpty && labelStyle == .default
-            ? .small
-            : 0
+        .small
+//        prefix.isEmpty && labelStyle == .default
```
sjavora, line 127: Commented code

sjavora (changes requested):

Sources/Orbit/Components/InputField.swift
PavelHolec: Thx, updated and used fixed padding

sjavora (approved):

PR #654, opened 2023-08-15
Update focus state for `InputField`

PavelHolec:
image


Sources/Orbit/Support/Forms/InputContent.swift
```diff
@@ -82,11 +89,21 @@ public struct InputContent<Content: View, Prefix: View, Suffix: View>: View {
         }
     }
 
+    private var focusOutlineColor: Color {
+        switch (message, isFocused) {
+            case (.error, true):        return .redNormal
+            case (.warning, true):      return .orangeNormal
+            case (.help, true):         return .blueNormal
+            case (_, true):             return .blueNormal
+            default:                    return .clear
```
sjavora, line 98: `case (_, false)`?

PavelHolec: Updated and added snapshot tests

sjavora (approved):

PR #678, opened 2023-09-01
Update `EmptyState` design, add newly extracted button stack builder

PavelHolec:
image

image


Sources/Orbit/Components/EmptyState.swift
```diff
@@ -65,30 +61,52 @@ public extension EmptyState {
...
+            buttons()
+        }
+    }
+
+    /// Creates Orbit EmptyState component with no action.
+    init(
+        _ title: String = "",
+        description: String = "",
+        illustration: Illustration.Image = .none
+    ) where Content == EmptyView, Buttons == EmptyView {
```
sjavora, line 84: Can we use a default value for `Buttons` in the init above instead? Or is that a problem because of the order of closures? If so, shouldn't buttons be first?

PavelHolec: Yep, the order of closures is the issue - even more that the Button when used in Content does not work with .status() modifier, but instead silently uses the Button as-is, as the custom content.

Another option would be to remove the custom content support completely, but this removes the flexibility for customization.

sjavora (approved):

PR #693, opened 2023-10-14
Extract illustrations to a separate target

### LINKED ISSUE #329: Extract illustrations to a separate target
Some clients might not want to bundle all illustrations in their apps, particularly in space-constrained cases such as app clips, where there is a 10 or 15 MB limit.

sjavora:
This should allow Orbit to be used in more space-constrained environments, like app clips.

`Orbit` target now builds in almost half the time (since some assets were moved to `OrbitIllustrations`):
| before | after |
| ------ | ----- |
| original | new |

This does affect snapshots, unfortunately - there is a new pink placeholder for illustrations, since the illustrations themselves are not availabie directly in Orbit anymore.

During implementation, I discovered that using `bundle: .module` works for the illustrations target. It seems like the original problem described in `Bundle.swift` (https://forums.swift.org/t/swift-package-with-resources-in-ios-project-wont-compile-test-targets/40846) has actually been fixed, so we should be able to now delete that file.

Closes #329


Automation/update_illustrations.py
```diff
@@ -43,6 +43,7 @@
 '''
 
 illustration_source_template = '''import Foundation
+import Orbit
```
PavelHolec, line 46: Improvement suggestion: Any idea for naming to improve the call site? We currently have it appear as an `Illustration."Image"` type that looks confusing ( = same name as native `Image` view type).

Maybe: `IllustrationImage` type? (We have `Symbol` for `Icon`)
 
image

PavelHolec (approved):
👍

Automation/update_illustrations.py
sjavora: You mean we would have `Illustration.IllustrationImage` or no nesting and just `IllustrationImage` at top level?

PavelHolec: Could be either. But in the nesting variant, it should be something like `Icon.Symbol` (`Illustration.Symbol`  would work for consistency, even if it may sound strange) that does not clash with a known SwiftUI name type.

PavelHolec: Another idea is `Illustration.Asset` (hinting about actual `xcasset` content that Illustrations use)

sjavora: Used `Illustration.Asset`

PavelHolec (approved):

PR #695, opened 2023-10-16
Unify background configuration components

PavelHolec:


Sources/Orbit/Support/ButtonStyles/TileButtonStyle.swift
```diff
@@ -0,0 +1,101 @@
+import SwiftUI
+
+/// Button style for Orbit  ``Tile`` or ``ChoiceTile`` components.
```
sjavora, line 3: "and" instead of "or" sounds better to me.

Sources/Orbit/Support/Environment Keys/BackgroundColorKey.swift
```diff
@@ -0,0 +1,68 @@
...
+            case let color as Color:            color
+            case let color as LinearGradient:   color
+            case let color as RadialGradient:   color
+            default:                            EmptyView()
+        }
+    }
+}
+
+struct BackgroundColorKey: EnvironmentKey {
+    static var defaultValue: BackgroundColor?
```
sjavora, line 29: `let`

Sources/Orbit/Support/Environment Keys/BackgroundColorKey.swift
```diff
@@ -0,0 +1,68 @@
...
+public extension EnvironmentValues {
+
+    /// A background color stored in a view’s environment, used for Orbit components that contain background, such as `Card` or `Tile`.
+    var backgroundColor: BackgroundColor? {
+        get { self[BackgroundColorKey.self] }
+        set { self[BackgroundColorKey.self] = newValue }
+    }
+}
+
+public extension View {
```
sjavora, line 41: Hmm... won't this be too similar to SwiftUI's existing `.background`?

Sources/Orbit/Support/Environment Keys/BackgroundColorKey.swift
```diff
@@ -0,0 +1,68 @@
...
+    func backgroundColor(_ shape: any ShapeStyle) -> some View {
+        environment(\.backgroundColor, .init(inactive: shape, active: shape))
+    }
+    
+    /// Set separate inactive and active background colors for any supported Orbit component within the view hierarchy.
+    ///
+    /// - Parameters:
+    ///   - shape: A color or a gradient that will be used in supported Orbit components such as `Card` or `Badge` as a background style.
+    ///   - active: A color or a gradient that will be used in supported touchable Orbit components such as `Tile` as an active (pressed) background style.
+    func backgroundColor(_ shape: any ShapeStyle, active: any ShapeStyle) -> some View {
```
sjavora, line 56: I saw this used in the first commit - since this one is totally standalone, you could just move it to be first instead.

Sources/Orbit/Components/Badge.swift
```diff
@@ -22,64 +24,68 @@ public struct Badge<LeadingIcon: View, TrailingIcon: View>: View, PotentiallyEmp
...
-        switch type {
-            case .light:                                Color.whiteDarker
-            case .lightInverted:                        Color.inkDark
-            case .neutral:                              Color.cloudLight
-            case .status(let status, true):             (status ?? defaultStatus).color
-            case .status(let status, false):            (status ?? defaultStatus).lightColor
-            case .custom(_, _, let backgroundColor):    backgroundColor
-            case .gradient(let gradient):               gradient.background
+    @ViewBuilder var resolvedBackground: some View {
+        if let backgroundColor = backgroundColor {
```
sjavora, line 51: We should really start using `SwiftFormat` in this project 🤔

Snapshots/iPhone/TagTests/testTags.1.png
sjavora, line 1: Seems like some colors got lost here...

sjavora (changes requested):

Sources/Orbit/Support/Environment Keys/BackgroundColorKey.swift
PavelHolec: This is a shorthand to not having to specify `= nil`

Sources/Orbit/Support/Environment Keys/BackgroundColorKey.swift
PavelHolec: It is, but I did not found a better name for it. It represents the `background` that would be used inside the component.
Another possibility would be `.backgroundShape`.

It is now matching with our `textColor` which is different to the `foregroundColor`.

Sources/Orbit/Support/Environment Keys/BackgroundColorKey.swift
PavelHolec: ✔️

Sources/Orbit/Support/Environment Keys/BackgroundColorKey.swift
PavelHolec: ✔️

Sources/Orbit/Components/Badge.swift
PavelHolec: ✔️

Snapshots/iPhone/TagTests/testTags.1.png
PavelHolec: Removed. The button styles could be internally hijacked to simulate the pressed state, but probably not worth at the moment (only Tags had this and it made the API confusing)

sjavora (approved):

PR #715, opened 2023-11-08
Concurrency safe Color extensions

Parabak:
Fixing warnings when `StrictConcurrency` is enabled

I've modified the script and run it to see the output.

### DIFF
--- a/Automation/update_colors.py
+++ b/Automation/update_colors.py
@@ -76,9 +76,9 @@ xcassets_color_template = '''{{
 source_filename_plural = f'Colors'
 source_group_template = '\n    // MARK: - {group}'
 
-source_line_template = '    /// Orbit {description} color.\n    static var {name} = Color("{description}", bundle: .orbit)'
+source_line_template = '    /// Orbit {description} color.\n    static let {name} = Color("{description}", bundle: .orbit)'
 source_line_template_shape = '    /// Orbit {description} shape style.\n    static var {name}: Self {{ .{name} }}'
-source_line_template_uicolor = '    /// Orbit {description} color.\n    static var {name} = fromResource(named: "{description}")'
+source_line_template_uicolor = '    /// Orbit {description} color.\n    static let {name} = fromResource(named: "{description}")'
 source_line_template_uiext = '            case .{name}: return.{name}'
 
 source_template = '''import SwiftUI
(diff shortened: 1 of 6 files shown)


Parabak:
I have to fix Storybook first

Parabak:
@PavelHolec We need to find solution for that.  `static var color` shares mutable state in different contexts and have to be fixed. In Swift6 it produces warning:

> Reference to static property 'orangeNormal' is not concurrency-safe because it involves shared mutable state

Simple `static let` will work but not for Storybook where we allow alternate colors 🤷 

I've tried to make it thread-safe:

```swift
extension DispatchQueue {

    static let atomicColor = DispatchQueue(
        label: "CustomColorSync",
        qos: .userInteractive,
        attributes: .concurrent
    )

}

@propertyWrapper struct Actor<Value: Sendable>: Sendable {

    private var _value: Value

    var wrappedValue: Value {
        get { DispatchQueue.atomicColor.sync { return _value } }
        set { DispatchQueue.atomicColor.sync { _value = newValue } }
    }

    init(wrappedValue: Value) {
        self._value = wrappedValue
    }
}
```
but that works 2 times slower on reading and 3 times slower on writing than normal access 🤷 It's too much for such case.

PavelHolec:
> @PavelHolec We need to find solution for that. `static var color` shares mutable state in different contexts and have to be fixed. In Swift6 it produces warning:
> 
> > Reference to static property 'orangeNormal' is not concurrency-safe because it involves shared mutable state
> 
> Simple `static let` will work but not for Storybook where we allow alternate colors 🤷
> 
> I've tried to make it thread-safe:
> 
> ```swift
> extension DispatchQueue {
> 
>     static let atomicColor = DispatchQueue(
>         label: "CustomColorSync",
>         qos: .userInteractive,
>         attributes: .concurrent
>     )
> 
> }
> 
> @propertyWrapper struct Actor<Value: Sendable>: Sendable {
> 
>     private var _value: Value
> 
>     var wrappedValue: Value {
>         get { DispatchQueue.atomicColor.sync { return _value } }
>         set { DispatchQueue.atomicColor.sync { _value = newValue } }
>     }
> 
>     init(wrappedValue: Value) {
>         self._value = wrappedValue
>     }
> }
> ```
> 
> but that works 2 times slower on reading and 3 times slower on writing than normal access 🤷 It's too much for such case.

In that case, let's drop the support to override colors in the storybook.

Parabak:
will update the MR

PavelHolec (approved):

PR #723, opened 2023-11-21
Added `identifier(_:AccessibilityID)` convenience modifier

PavelHolec:
Allowing call site to specify AccessibilityID also as an Identifier using a single modifier at the same time.


sjavora:
Seems like an antipattern to me 🤷. If we have to have this, why not put it in the app instead?

PavelHolec:
Closed, but kept the branch for future reference

PR #747, opened 2024-01-10
Update `Card` component design and action API

PavelHolec:
image


Sources/OrbitStorybook/TutorialScreen.swift
```diff
@@ -27,6 +27,8 @@ struct TutorialScreen: View {
                     Button("Secondary Button", type: .secondary, action: secondaryAction)
                     Button(icon: .creditCard, type: .status(.warning), action: {})
                 }
+            } action: {
+                ButtonLink("Edit", action: {})
```
sjavora, line 31: Will this ever be something else than `ButtonLink`? I'm assuming you think so and that's why this is a generic `@ViewBuilder` as opposed to something like we have for alerts/dialogs, right?

sjavora (approved):

Sources/OrbitStorybook/TutorialScreen.swift
PavelHolec: Yes, this container serves both use-cases: button(link) action (primary usage) and a generic trailing header placeholder (for small illustrations etc.). A `header` name would be better, but that might be confusing for the primary usage.

PR #759, opened 2024-01-29
Update `Switch` design

PavelHolec:


sjavora (approved):
Wow, so now we're basically back to the system version, except for an optional icon?

PR #772, opened 2024-03-01
Update `Card` to use environment value to drive the content layout

PavelHolec:
This combines previous flexibility with the ability to configure multiple Cards via environment.


Sources/Orbit/Components/Card.swift
```diff
@@ -192,6 +186,39 @@ public extension AccessibilityID {
...
+// MARK: - Types
+
+/// The layout of Orbit ``Card`` stack content.
+public enum CardLayout {
+    
+    /// Content layout uses a `VStack` with customizable padding and spacing.
+    case custom(alignment: HorizontalAlignment = .leading, padding: CGFloat = .medium, spacing: CGFloat = .medium)
+    
+    /// Content layout uses a `VStack` with default `.medium` padding and spacing.
+    public static var `default` = Self.custom()
```
sjavora, line 198: `let`

Sources/Orbit/Components/Card.swift
```diff
@@ -192,6 +186,39 @@ public extension AccessibilityID {
     static let cardAction               = Self(rawValue: "orbit.card.action")
 }
 
+// MARK: - Types
+
+/// The layout of Orbit ``Card`` stack content.
+public enum CardLayout {
+    
+    /// Content layout uses a `VStack` with customizable padding and spacing.
+    case custom(alignment: HorizontalAlignment = .leading, padding: CGFloat = .medium, spacing: CGFloat = .medium)
```
sjavora, line 195: Do you anticipate more `case`s here? If not, there isn't really a reason for this to be an `enum`.

sjavora (approved):

Sources/Orbit/Components/Card.swift
PavelHolec: I was thinking about future cases like `lazyStack` (which would use iOS14+ LazyVStack for both stacks that are currently used) etc. But it makes sense to keep it as a struct for now.

Sources/Orbit/Components/Card.swift
PavelHolec: Updated

Sources/Orbit/Components/Card.swift
PavelHolec: Replaced with a struct with two public predefined values (`default` and `fill`)

sjavora (approved):

Issue #780, opened 2024-03-14
Add localization support for components


PavelHolec:
With the localizable `Text`, add support to all other components.

PR #781, opened 2024-03-14
Add localization support to Text component

PavelHolec:
The components update will follow in #780

### DIFF
--- a/Sources/Orbit/Components/Text.swift
+++ b/Sources/Orbit/Components/Text.swift
@@ -1,9 +1,10 @@
 import SwiftUI
 
 /// Orbit component that displays one or more lines of read-only text. 
-/// A counterpart of the native `SwiftUI.Text` with a support for Orbit markup formatting and ``TextLink``s.
+/// A counterpart of the native `SwiftUI.Text` with a support for Orbit markup formatting, ``TextLink``s and custom vertical padding.
 ///
-/// A ``Text`` is created using the `String` content that can include html tags `<strong>`, `<u>`, `<ref>`, `<br>` to customize formatting. 
+/// A ``Text`` is created using the `String` verbatim or localizable content that can include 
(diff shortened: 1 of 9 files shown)


Sources/Orbit/Components/Text.swift
```diff
@@ -304,18 +345,83 @@ public struct Text: View, FormattedTextBuildable, PotentiallyEmptyView {
...
+    @_semantics("swiftui.init_with_localization")
+    init(
+        _ key: StaticString,
+        value: LocalizedStringKey,
+        tableName: String? = nil,
+        bundle: Bundle? = nil,
+        comment: StaticString? = nil
+    ) {
+        self.localization = .key(value, bundle: bundle, tableName: tableName, explicitKey: key.description)
+        self.verbatimContent = value.key ?? ""
```
sjavora, line 370: Shouldn't this fall back to `key`?

Sources/Orbit/Support/Text/LocalizedStringKey.swift
```diff
@@ -0,0 +1,84 @@
...
+                forResource: locale.identifier.split(separator: "_").map(String.init).first ?? locale.identifier, 
+                ofType: "lproj"
+            )
+        return (bundlePath.map(Bundle.init(path:)) ?? self) ?? self
+    }
+}
+
+public extension LocalizedStringKey {
+
+    private typealias FArgs = (CVarArg, Formatter?)
```
sjavora, line 17: `FormatterArguments`? I really can't tell what this short name is supposed to be... 😅

PavelHolec: Renamed to `FormattedVarArgs`

Sources/Orbit/Components/Text.swift
PavelHolec: Updated

sjavora (approved):

PR #783, opened 2024-03-18
Add localization support for components

PavelHolec:
The task required to replace all `String` properties with a custom content.


Sources/Orbit/Components/KeyValue.swift
```diff
@@ -15,76 +15,85 @@ import SwiftUI
...
-                Text(value)
-                    .textSize(size.valueSize)
-                    .fontWeight(.medium)
+            VStack(alignment: .init(multilineTextAlignment), spacing: 0) {
+                key
+                    .textColor(.inkNormal)
+
+                value
+                    .textFontWeight(.medium)
+                    .textSize(.large)
```
sjavora, line 33: This is also on the VStack.

Sources/Orbit/Components/KeyValue.swift
```diff
@@ -15,76 +15,85 @@ import SwiftUI
...
-extension KeyValue {
-
-    /// Creates Orbit ``KeyValue`` component.
+    
+    /// Creates Orbit ``KeyValue`` component with custom content.
     public init(
-        _ key: String = "",
-        value: String = "",
-        size: Size = .normal
+        @ViewBuilder value: () -> Value,
```
sjavora, line 53: Why is `value` first?

Sources/Orbit/Components/Badge.swift
```diff
@@ -32,17 +32,17 @@ import SwiftUI
...
     @Environment(\.textColor) private var textColor
 
-    private let label: String
-    private let leadingIcon: LeadingIcon
-    private let trailingIcon: TrailingIcon
-    private let type: BadgeType
+    @ViewBuilder private let leadingIcon: LeadingIcon
+    @ViewBuilder private let label: Label
+    @ViewBuilder private let trailingIcon: TrailingIcon
+    @ViewBuilder private let type: BadgeType
```
sjavora, line 45: Is there a point to `@ViewBuilder` here if it's also used in the public init?

Sources/Orbit/Components/Collapse.swift
```diff
@@ -11,97 +11,155 @@ import SwiftUI
...
-                        header
-                            .accessibility(.collapseHeader)
-
-                        Spacer(minLength: .xSmall)
-
-                        Icon(.chevronDown)
-                            .rotationEffect(isExpanded ? .degrees(180) : .zero)
-                    }
-                    .padding(.vertical, headerVerticalPadding)
+		OptionalBinding(isExpanded) { $isExpanded in
```
sjavora, line 22: Alignment

Sources/Orbit/Components/SegmentedSwitch.swift
```diff
@@ -173,6 +176,10 @@ public struct SegmentedSwitch<Selection: Hashable, Content: View>: View {
         idealSize.horizontal == true
     }
     
+    private var isSelected: Bool {
+        (selection as AnyObject) is NSNull == false
```
sjavora, line 180: Could you explain this one?

Sources/Orbit/Support/Forms/SelectableLabelWrapper.swift
```diff
@@ -30,8 +30,10 @@ private final class CopyableLabel: UILabel {
         super.init(frame: frame)
 
         isUserInteractionEnabled = true
-        isAccessibilityElement = true
-        accessibilityTraits = .staticText
+        
+        isAccessibilityElement = false
+//        accessibilityTraits = .staticText
```
sjavora, line 35: Commented code

Sources/Orbit/Components/KeyValue.swift
PavelHolec: This mimics native component custom viewbuilder inits, where (with exceptions) the order of parameters is switched (compared to variants where the non-viewbuilder parameters), so that the primary content or action associated with the component uses the first unnamed closure, while the rest uses closure labels.

Sources/Orbit/Components/Badge.swift
PavelHolec: It just makes it easier to distinguish "subcomponent views" from other non-view properties on a first glance. 

It would be technically required, but only in case some of our inits used viewbuilder syntax to initialize the content manually based on init parameters.

Sources/Orbit/Components/SegmentedSwitch.swift
PavelHolec: This obscure workaround came from [here](https://forums.swift.org/t/how-to-check-if-a-value-of-a-generic-type-is-nil/50631/10).

Sources/Orbit/Components/KeyValue.swift
PavelHolec: Thx, fixed

Sources/Orbit/Components/Collapse.swift
PavelHolec: Fixed

Sources/Orbit/Components/SegmentedSwitch.swift
sjavora: Let's add a comment about it then 👍

PavelHolec: Updated

sjavora (approved):

PR #794, opened 2024-04-12
Fix tag remove action

PavelHolec:
The explicit flag makes the API less awkward to use remove action.


sjavora:
Is this necessary? It makes weird invalid combinations possible...

sjavora (approved):

PR #855, opened 2024-11-12
Fix snapshot test regeneration job

PavelHolec:


Sources/Orbit/Support/BindingSource.swift
```diff
@@ -6,6 +6,8 @@ public enum OptionalBindingSource<Value> {
     case state(Value)
 }
 
+extension Binding: Sendable where Value: Sendable {}
```
sjavora, line 9: Why is this necessary? `Binding` already conforms to `Sendable`. Is this for some older OSes?

Sources/Orbit/Support/Previews/PreviewWrapper.swift
```diff
@@ -1,7 +1,7 @@
 import SwiftUI
 
 /// Registers Orbit fonts for preview content.
-public struct OrbitPreviewWrapper<Content: View>: View {
+@MainActor public struct OrbitPreviewWrapper<Content: View>: View {
```
sjavora, line 4: `View` is already `@MainActor` 🤔

Sources/Orbit/Support/BindingSource.swift
PavelHolec: Good point. This plus below were errors in my previous attempt to fix snapshot testing builds. It may not be necessary anymore, will check...

PavelHolec: Removed

Sources/Orbit/Support/Previews/PreviewWrapper.swift
PavelHolec: Removed

sjavora (approved):

